PR #6, opened 2016-04-11
Add spotify-heroic-datasource version 0.0.1

udoprog:


bergquist:
@udoprog Seems like you forgot to checkin the dist folder in that repo. We need the "compiled" files as well since we dont want to run build scripts when installing plugins :)

bergquist:
All new plugins id's should follow this pattern <org/username>-<name>-<type> 

So in this case: spotify-heroic-datasource

We have not implemented automatic validation yet so we have to act like validation robots our self at the moment :)

udoprog:
@bergquist The referenced commit-id contains the built code to avoid poluting the main repo. Is this OK?
https://github.com/udoprog/grafana-heroic-datasource/tree/62f8f8b3ce41b6649d8fb874ab7c0836b4618cb2

As for the id, what exactly should the first part reference? This plugin was written on my personal account and time and is unaffiliated with Spotify.

udoprog:
I've updated the repo, and pull request to use the correct id `spotify-heroic-datasource`.

Also, keep in mind that the built code again lives in a separate commit:
https://github.com/udoprog/spotify-heroic-datasource/commit/d662044751a93031351ce588a6227240bc7b88a9

Looking through the grafana-cli utility, I think this should work, but please advise.

bergquist:
@udoprog either udoprod-heroic-datasource or spotify-heroic-datasource. Thats your decision :)

Ah. That commit looks much better :) Sorry about that. 

Would be awesome if it contained "heroic-screenshot1.png" since it would showup on grafana.net plugin page like this https://grafana.net/plugins/alexanderzobnin-zabbix-app. Would be nice to link it in the readme as well.

bergquist:
@udoprog nice! Should show up on grafana.net within a few min :)

PR #9, opened 2016-04-12
Add udoprog-heroic-datasource version 0.0.1

udoprog:
Hey,

So this should introduce udoprog-heroic-datasource, hopefully all good this time.
One thing I noticed last time when attempting to install the previous plugin from grafana.net is that the partials didn't download correctly. Building locally and symlinking the directory to `$GF_PLUGIN_DIR/heroic` works though.

I've created an account on grafana.net now as well.

### DIFF
--- a/repo.json
+++ b/repo.json
@@ -173,6 +173,18 @@
           "url": "https://github.com/raintank/snap-app"
         }
       ]
+    },
+    {
+      "id": "udoprog-heroic-datasource",
+      "type": "datasource",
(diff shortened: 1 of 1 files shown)


udoprog:
Ah. I realised it might just have been me picking graphite instead as a source by accident. Carry on :)

bergquist:
@udoprog The plugin works fine when downloading the linked version and placing it in data/plugins. 

@DanCech considering yesterdays hassle I think you should merge this in case we have to un-publish it again :)

DanCech:
Looks great!

PR #17, opened 2016-05-11
Release v2.0.1 of the OpenNMS Datasource plugin.

j-white:
The plugin was updated to use the new query editor styles.


daniellee:
Merged and pushed to Grafana.net.

PR #18, opened 2016-05-19
Add udoprog-heroic-datasource version 0.1.0 (replace 0.0.1)

udoprog:
There was an incompatible change from one of the Beta versions which made the input field break (see https://github.com/udoprog/udoprog-heroic-datasource/pull/2), so there's no good reason to keep version `0.0.1` around.

### DIFF
--- a/repo.json
+++ b/repo.json
@@ -316,6 +316,11 @@
           "version": "0.0.1",
           "commit": "4c922cee6d8127c87bd0f4b7e1e4c4fa22ff2c91",
           "url": "https://github.com/udoprog/udoprog-heroic-datasource"
+        },
+        {
+          "version": "0.1.0",
+          "commit": "ebe7dffb02b0843b34b043786f651ad02205c417",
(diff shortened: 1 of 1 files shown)


DanCech:
Our general policy is not to remove older versions of the plugins, can you update the PR to add 0.1.0 as a new version instead?

udoprog:
@DanCech fixed

DanCech:
Great! Now there are a couple other things you'll need to do to make the plugin install-able via grafana-cli.

1. Update plugin.json with the correct `version` and `updated` date
2. Add the `dist` folder produced by `grunt` to the repo

udoprog:
Hey, the commit being references is [this](https://github.com/udoprog/udoprog-heroic-datasource/commit/ebe7dffb02b0843b34b043786f651ad02205c417), it contains everything you mentioned but separate from the uncompiled source.

DanCech:
Aha, ok.

DanCech:
version 0.1.0 is now live https://grafana.net/plugins/udoprog-heroic-datasource

PR #21, opened 2016-06-06
Add stagemonitor-grafana-elasticsearch app

felixbarny:
0.26.0 is not released yet, thus the commit hash is empty.

If you think the https://github.com/stagemonitor/stagemonitor-grafana-elasticsearch looks good, I'll release the final version.

I've already created a stagemonitor organization on grafana.net.

### DIFF
--- a/repo.json
+++ b/repo.json
@@ -372,6 +372,18 @@
           "url": "https://github.com/percona/grafana-app"
         }
       ]
+    },
+    {
+      "id": "stagemonitor-grafana-elasticsearch",
+      "type": "app",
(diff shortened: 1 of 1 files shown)


DanCech:
Thanks for the PR @felixbarny 

There are a few things we'll need to tweak before it's ready to go onto grafana.net.  Having your org set up is a good start!  You will need to change the plugin `id` to match our naming convention, in this case I'd suggest `stagemonitor-elasticsearch-app`.

You will also need to have a valid plugin.json in the root of your plugin, with correct paths to the logos & screenshots so grafana.net can pick them up.

PR #22, opened 2016-06-20
add histogram/heatmap panel

mtanda:


daniellee:
Hi @mtanda great to see these two panels come in as a PR! 

One minor thing needs to be changed. The id, both in this repo and in the plugin json needs to match the naming conventions for plugins. This means that the ids **grafana-heatmap-epoch-panel** and **grafana-histogram-panel** should be changed to **mtanda-heatmap-epoch-panel** and **mtanda-histogram-panel** (I am assuming you want it under your own name?)

daniellee:
Also, the path for the image in heatmap epoch panel (and the histogram panel) is incorrect. If there is a dist folder, then Grafana will use that first and then the path `src/images/heatmap.png` should be 'images/heatmap.png'. You don't have to have a plugin.json in the root so you could move it into the src folder like [this](https://github.com/raintank/ns1-app) plugin does. 

Sorry I know it is a bit messy with paths.

mtanda:
@daniellee Ah, I misunderstand naming convention :-)
I fix the plugin name and image path.

DanCech:
Looking great!

mtanda:
Thanks!!

PR #23, opened 2016-06-25
Add grafana-sunandmoon-datasource plugin (v0.1.0)

fetzerch:
I've been working on a small datasource plugin, that calculates the sun and moon position - useful, if you use Grafana for weather station projects.

Screenshot

Since it's my first JS project, a code review would be highly appreciated.

I'm using Grafana for my [atMETEO](https://fetzerch.github.io/2015/05/20/atmeteo/) project and this is how I make use of the datasource in a bigger dashboard. It's useful to see for example the current moon phase or sunrise / sunset times.
atMETEO grafana

I'd be happy if this could get added to grafana.net.
Cheers,
Christian

### DIFF
--- a/repo.json
+++ b/repo.json
@@ -455,6 +455,18 @@
           "url": "https://github.com/mtanda/grafana-heatmap-epoch-panel"
         }
       ]
+    },
+    {
+      "id": "fetzerch-sunandmoon-datasource",
+      "type": "datasource",
+      "url": "https://github.com/fetzerch/grafana-sunandmoon-datasource",
+      "versions": [
+        {
+          "version": "0.1.0",
+          "commit": "b4cb66988aca9b59576b15e762c5fcda056bde08",
+          "url": "https://github.com/fetzerch/grafana-sunandmoon-datasource"
+        }
+      ]
     }
   ]
 }


fetzerch:
Updated the PR to not use the `grafana-` prefix as plugin id. Let me know if there are further changes needed.

daniellee:
Oh wow, thanks for sending this in! This looks amazing. Before we review the actual plugin, can you fix the path for the screenshot: `"path": "img/screenshot.png"` needs to have src added to the path.

fetzerch:
Thanks, done.

One thing I noticed is that I can't enable multiple annotations. In the screenshot above if I enable also "Noon", it would still only show the annotations for "Sunrise / Sunset". If I only enable "Noon" it works correctly. I'm not sure if that's the correct behavior or if I did something wrong in the plugin. (That's with grafana 3.1beta).

daniellee:
I have been showing your data source to everyone working at Raintank (the company behind Grafana) and they were all really impressed. You asked for a code review and I think the JavaScript generally looks fine. I have some feedback here: 

1. We had a bit of debate on whether the location settings should be under the data source settings or if they should be in a new dashboard settings tab. I think it is fine as it is but if you want to add more settings then maybe you should consider moving them into the dashboard settings.

2. I sent you a [small PR](https://github.com/fetzerch/grafana-sunandmoon-datasource/pull/1) that fixes some small stuff.

3. Yes, there is something broken there with multiple annotations. It should work. I will get back to you later today when I've figured out if it is a bug in Grafana or something that we can fix in your code.

daniellee:
Torkel just demoed your data source at Monitorama in Portland! Going to publish it now and we'll see if we can fix the annotations later today.

daniellee:
Published here: [https://grafana.net/plugins/fetzerch-sunandmoon-datasource](https://grafana.net/plugins/fetzerch-sunandmoon-datasource)

fetzerch:
Thanks for the positive feedback! :). As a long-time grafana user, it's a good opportunity to give something back to the community. I'm really happy that you like it!

1. Sure, I can look into making it a dashboard setting. I completely forgot about that possibility. One thing that is certainly nice to support is being able to use multiple locations. I'm thinking for example that one might want to monitor a farm of photovoltaic systems around the world. Is this possible with dashboard settings? Guess they are per dashboard, right?

2 & 3. Thanks :)

I can also share ideas for future features:
- For photovoltaic systems the clear sky radiation could be calculated as well
- The annotations could be extended to be able to show an annotation at midnight and noon (or at certain hours). Maybe that would better fit into a separate addon or the grafana core (maybe even together with annotation ranges).
- Annotations for [solstices, equinoxes, perihelion, aphelion](https://github.com/mourner/suncalc/issues/60)

daniellee:
[PR sent](https://github.com/fetzerch/grafana-sunandmoon-datasource/pulls) for the multiple annotations bug.

Sorry, I didn't mean per dashboard, I meant per panel. You could add a settings tab per panel. There is no way to do it per dashboard. Then for your example, you could have one panel per farm. Would that work for you?

fetzerch:
Thanks!

Yes that sounds good. To be honest for my main use case, I only use it in one panel with my home location. I'll try to look into it in the next days.

fetzerch:
@daniellee: I started with panel options (in https://github.com/fetzerch/grafana-sunandmoon-datasource/commit/f15ae4a15d9a28220057bbd763027f0479ecd91c) but ran into a few issues. Would be great if you could assist.

- What do I have to set `ng-model` to? All the current existing datasources use a set of the three values: `interval`, `cacheTimeout` and `maxDataPoints` and they are 'hardcoded' in https://github.com/grafana/grafana/blob/master/public/app/features/panel/metrics_panel_ctrl.ts#L179, so I cannot add my vars to the `panel` object. Do I need to use `scopedVars` for this somehow?
- With the current way of using datasource settings, the configuration values can easily be verified with the `testDatasource` method. Is there something similar for panel options?

I've tried to get that information by looking into the other existing plugins, but I could not find one that uses the panel options in such a way. Thanks for your help!

PR #31, opened 2016-07-07
update heatmap panel

mtanda:
add legend support

### DIFF
--- a/repo.json
+++ b/repo.json
@@ -475,6 +475,31 @@
       "url": "https://github.com/mtanda/grafana-heatmap-epoch-panel",
       "versions": [
         {
+          "version": "0.1.7",
+          "commit": "e1ec956e86e55d9a3d06554df03641773c146fa9",
+          "url": "https://github.com/mtanda/grafana-heatmap-epoch-panel"
+        },
+        {
+          "version": "0.1.6",
+          "commit": "84d37548d8b3e1a1ba73bafe28059395c2eac68d",
+          "url": "https://github.com/mtanda/grafana-heatmap-epoch-panel"
+        },
+        {
+          "version": "0.1.5",
+          "commit": "8ee7d63b71f10700c2bbda508c2411b68f06fe29",
+          "url": "https://github.com/mtanda/grafana-heatmap-epoch-panel"
+        },
+        {
+          "version": "0.1.4",
+          "commit": "caea4c023586702c56419c15e1798a211cd35caf",
+          "url": "https://github.com/mtanda/grafana-heatmap-epoch-panel"
+        },
+        {
+          "version": "0.1.3",
+          "commit": "6144fac8205a709c1d406c90e21f84e6a8f30254",
+          "url": "https://github.com/mtanda/grafana-heatmap-epoch-panel"
+        },
+        {
           "version": "0.1.2",
           "commit": "f1a15ae5bdcea5ad6c26ed6b663171b0937c3207",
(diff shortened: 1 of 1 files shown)


daniellee:
Hi @mtanda I'm still seeing some graphical glitches:

image

- The legend when shown to the right is cut off.
- The color for the x and y axis is too dark, it should be the same as the histogram.
- When refreshing data after making a change to the metric query, the panel often does not get updated.
- I saw this error on the console: "heatmap epoch panel warning: exceed max time series (support5 time series)". Is there a better way of showing this error?
- The dates on the x-axis sometimes get all clumped together.

I'm seeing some other problems but I will try and find the causes and send in detailed issues for them.

mtanda:
@daniellee Thanks for feedback.

> The legend when shown to the right is cut off.

I don't try it yet, thanks for find this bug.

> The color for the x and y axis is too dark, it should be the same as the histogram.

The color is defined in Epoch's CSS, and I can't find the way to change the color...

> When refreshing data after making a change to the metric query, the panel often does not get updated.

The panel plugin reuse Epoch instance, it sometimes behave badly.
If you have the replicatable dashboard, sharing the snapshot is great help for me :-)

> I saw this error on the console: "heatmap epoch panel warning: exceed max time series (support5 time series)". Is there a better way of showing this error?

This is the limitation of Epoch plugin. Maybe I should show warning on Grafana, and explain the limitation in README.

> The dates on the x-axis sometimes get all clumped together.

I'll try to replicate the bug.

mtanda:
@daniellee I fixed the issue. Would you check latest version?

mtanda:
Fix timezone handling, update.

mtanda:
Does anyone check this? (cc: @torkelo)

torkelo:
we are low staffed this week, should be better next week :)

mtanda:
hmm, okay...

mtanda:
@torkelo When the review is started...? (I'm sorry for my impatience.)

daniellee:
Hi @mtanda sorry for the delay, I was on vacation. I tested the latest version and I am experiencing two problems. 

1. I can't save a dashboard with a heatmap epoch - I get the following error: 

> TypeError: Converting circular structure to JSON

To reproduce: create a new dashboard, add a heatmap epoch panel and then try to save.

2. I had a heatmap epoch panel that was created with a previous version, it now hangs both Chrome and Firefox.

I don't have much time next week but after that I will try and help you out more. We would love to get this working with Grafana!

mtanda:
Hi, I'm sorry I don't know you are on vacation...

I check the reported problems, and I success to replicate "circular structure" error. @ https://github.com/mtanda/grafana-heatmap-epoch-panel/commit/4ace08a4bd7b454c63d9e705fda1f8050a17a8b2 .
The "obj.dashboard.rows[0].panels[0].heatmapOptions.el" has circular structure.
I can't save the dashboard due to the error.

So I just upgrade heatmap epoch panel to latest version, and do same thing, the error has gone, doesn't occur.

The DOM element has problematic reference cause circular reference.
Plugin could check the problematic setting, and remove it.
But, re-create the heatmap panel is safe way, I think.

mtanda:
I think this PR never merged...

mtanda:
😞

mtanda:
Thanks for updating grafana.net.
https://grafana.net/plugins/mtanda-heatmap-epoch-panel

PR #35, opened 2016-08-03
Add LatestCommitId

abhisant:
Add the commit-id which has the version number in plugin.json

### DIFF
--- a/repo.json
+++ b/repo.json
@@ -534,7 +534,7 @@
       "versions": [
         {
           "version": "0.0.1",
-          "commit": "2395d13961fc6c9755073574f62ba47d4a099c46",
+          "commit": "b0395aea1efa7e390319736422ba584c6d5cd74d",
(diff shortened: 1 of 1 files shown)


DanCech:
Thanks for that, it's merged and published.  One thing I do note is that there is no `README.md` under the `dist/` folder (only in the root), which means that the readme content won't show up on grafana.net or in grafana itself.  You may want to create `dist/README.md` (and release a new version) so that the readme will show up properly.

abhisant:
Yes, I was about to mention that. Should I raise a new quick PR. In the same version ?

abhisant:
Created a pull request: https://github.com/grafana/grafana-plugin-repository/pull/37 <https://github.com/grafana/grafana-plugin-repository/pull/37>
Having a higher version (0.0.2) and README.md in dist.

PR #48, opened 2016-10-07
updated commit version to include new icon and sequence diagram css fix

jdbranham:


DanCech:
In general we don't change released versions once they've been published, you're welcome to release 1.2.2 though

DanCech:
Ah, I see 1.2.1 was broken anyway, I'll make an exception this time

jdbranham:
cool, thanks Dan!
I'll double check the plugin.json next time!

DanCech:
I'd also suggest tweaking the headings in your README.md to have a space between the hashes and the text, github will render them as headings but the markdown parser we use (remarkable) doesn't recognize it without the space.

jdbranham:
thanks. will do.

PR #51, opened 2016-10-13
added grafana-diagram v1.3.0

jdbranham:


daniellee:
Merged and published. I see you're working hard on this, loads of releases the last few weeks! And lots of issues too (which is a good, means people are using it).

PR #52, opened 2016-10-14
Add Foursquare's Cloudera Manager datasource to plugins list

tdyas:


tdyas:
Questions:

1. Is the info in the plugin's `plugin.json` sufficient?

2. Does the ID have to be consistent between the `repo.json` and a plugin's `plugin.json`?

DanCech:
You'll want to flesh `plugin.json` out a bit more, and the `id` in `plugin.json` does need to match what's in the repo.  This is a good example https://github.com/jdbranham/grafana-diagram/blob/master/src/plugin.json

If you check out that repo you'll see that it has `plugin.json` in the `src` folder, which is now the preferred layout and requires only small changes to `Gruntfile.js`.  You'll also see that the built `dist` folder is checked in, which is required for the plugin to be install-able via `grafana-cli`.

I'd also recommend updating `README.md` since once the plugin is published on grafana.net the simplest way to install it will be via `grafana-cli`.

daniellee:
@tdyas Thanks for sending in this plugin. I don't have access to a Cloudera system but from what I could see it looks fine. There's not much that can go wrong when the query editor is so simple :-) I liked that you had links to the query language to help someone starting out.

The only problem is that in the dist folder the README.md filename is not uppercased. The convention is that it is always uppercased and it is a requirement for our import tool. If you can fix that then we can publish this straight away.

tdyas:
I fixed the README.md to be uppercased and filled in the recommended metadata. The updated commit points at the 0.9.2 version.

daniellee:
Merged and published: https://grafana.net/plugins/foursquare-clouderamanager-datasource

Thanks for submitting this!

PR #58, opened 2016-11-08
Add neocat-cal-heatmap-panel entry

NeoCat:
This plugin provides a calendar view with heatmap, like GitHub contribution calendar.

### DIFF
--- a/repo.json
+++ b/repo.json
@@ -778,6 +778,18 @@
           "url": "https://github.com/ryantxu/ajax-panel"
         }
       ]
+    },
+    {
+      "id": "neocat-cal-heatmap-panel",
+      "type": "panel",
+      "url": "https://github.com/NeoCat/grafana-cal-heatmap-panel",
+      "versions": [
+        {
+          "version": "0.0.2",
+          "commit": "edff2df73d699edb27f9bb3ec1f53a22fe4fc04d",
+          "url": "https://github.com/NeoCat/grafana-cal-heatmap-panel"
+        }
+      ]
     }
   ]
 }


daniellee:
Thanks for submitting this! The screenshots look amazing. I'll get back to you with feedback in a day or two. Don't worry about the conflict, I can fix that.

daniellee:
Hi @NeoCat 

This plugin looks great! This will be a fantastic addition to the plugin catalog.

No problems with the meta data (plugin.json and README.md), everything looks correct there.

## Feedback on the plugin

- The plugin crashes due to an uncaught exception if you choose the wrong option in the Sub domain dropdown. The page has to be refreshed to get it to work again. For example, if I choose `hour` in the Domain dropdown and `week` in the Sub domain dropdown (which is not allowed by cal-heatmap), it crashes with the following error:

image
 
And after changing both dropdowns back to Auto, the panel looks like this:

image

Maybe, you could hide options in the Sub domain dropdown depending on what is chosen in the Domain dropdown, or use validation to not allow an invalid choice.

- The options for cal-heatmap aren't very obvious as a first time user. A link to the documentation would be handy, maybe in an info toggle:

image

 or in a tooltip:

image

- The `Legend` field in my opinion should be called `Thresholds` even if it doesn't work the same way as thresholds for the graph panel. The values there affect the whole panel not just the legend.

If you fix the Domain/Sub Domain crash problem, then I can publish this straight away. Thanks again for submitting this!

Cheers,
Daniel

NeoCat:
Thank you for detailed feedback!

I have update the plugin and this plugin catalog entry:

(1) The Domain/SubDomain crash problem has been fixed. Now the domains / subDomains dropdown are limited to valid combinations, so that it will not crashed by invalid combinations.

(2) Tooltip to describe some of cal-heatmap options are added.

(3) Now 'Legend' has been renamed to 'Thresholds', and added an example in the tooltip.

(4) version is bumped to '0.0.2'.

Thanks.

daniellee:
Great work! Merged and published: https://grafana.net/plugins/neocat-cal-heatmap-panel

PR #59, opened 2016-11-08
added diagram v1.3.1

jdbranham:


daniellee:
Published. Nice work! One little suggestion -> could you add a refresh after changing a threshold so that the colors update on the diagram immediately.

PR #63, opened 2016-11-11
Add install instructions

timyllym:


daniellee:
We publish the README.md from the dist directory (and that's the one that is shown in Grafana too), can you run grunt so that it gets updated too. Thanks.

daniellee:
Thanks Timo. Merged and published: https://grafana.net/plugins/digiapulssi-breadcrumb-panel

PR #70, opened 2016-12-09
update Belugacdn app to 1.1.1

140am:
- update company logo

https://github.com/belugacdn/grafana-belugacdn-app/compare/v1.1.0...v1.1.1

### DIFF
--- a/repo.json
+++ b/repo.json
@@ -865,8 +865,8 @@
       "url": "https://github.com/belugacdn/grafana-belugacdn-app",
       "versions": [
         {
-          "version": "1.1.0",
-          "commit": "59022c2aa28c7bbcee1d7fca19fedebf786d18bd",
+          "version": "1.1.1",
+          "commit": "6cb4392e6877584b17642445bc392f9bb7117ddb",
           "url": "https://github.com/belugacdn/grafana-belugacdn-app"
         }
       ]
--- a/repo.json
+++ b/repo.json
@@ -865,6 +865,16 @@
       "url": "https://github.com/belugacdn/grafana-belugacdn-app",
       "versions": [
         {
+          "version": "1.0.0",
+          "commit": "979b635b7450a4f6ea55062766792dcb50c72abd",
+          "url": "https://github.com/belugacdn/grafana-belugacdn-app"
+        },
+        {
+          "version": "1.1.0",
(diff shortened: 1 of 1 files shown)


DanCech:
Please add the new version above the old one rather than replacing it.

140am:
@DanCech I have added the update as a new `versions` entry and also added the original 1.0.0 reference from https://github.com/grafana/grafana-plugin-repository/pull/69.

DanCech:
You need to update the version in plugin.json to 1.1.1 before grafana-net will accept it

140am:
Forgot about that :( This will result in a new commit hash - should i open a pull request for an updated 1.1.1 `versions` entry afterwards or tag it as 1.1.2 ?

daniellee:
Hi @140am 

I messed up when publishing this at GrafanaCon and missed that the plugin id is incorrect. Currently it is  `grafana-belugacdn-app'  but it should be just 'belugacdn-app'. The grafana part means the app got published under the Grafana org as the convention for the plugin id is org-name-type or org-type.

This is going to be a problem for people who have already installed the app. When the plugin id is changed they won't get any upgrade notifications when you release a new version. 40-50 users have it installed currently according to our stats.

I really apologise for this and for any inconvenience to your users. Do you have any way of contacting them easily? Our stats are anonymous so we only know how many not who.

Regards,
Daniel

140am:
@daniellee updating the package.json right now with the new `id` and will create a pull request under a 1.2.0 tag shortly. is it ok if i remove/replace the just added 1.1.1 tag with the new 1.2.0 as this only introduced the logo update or prefer a new addition and keep the 1.1.1 entry? we unfortunately do not have a way to contact users specifically but will emphasis the new installation ID on upcoming future announcements.

daniellee:
It's fine to remove 1.1.1 as it is not published.

Let us know if we can do anything to help - maybe a tweet from the Grafana account?

140am:
@daniellee i created a new pull request with the requested changes.

thank you for the generous offer and prompt help - we would love a tweet from your side about our plugin. we are working on some announcements around using grafana together with our plugin for infrastructure monitoring where we will mention the new plugin id. once we will do a bigger plugin update we will highlight that a new installation with `belugacdn-app` as id will be required to address existing users.

PR #71, opened 2016-12-09
Update Belugacdn app to 1.2.0

140am:
- this **changes the `id`** of the plugin as per https://github.com/grafana/grafana-plugin-repository/pull/70
- changes order of `versions` entry to have new versions above the old ones
- replaces just published 1.1.1 with 1.2.0 tag to address `id` change and missing plugin.json version update mentioned in https://github.com/grafana/grafana-plugin-repository/pull/70

### DIFF
--- a/repo.json
+++ b/repo.json
@@ -865,8 +865,8 @@
       "url": "https://github.com/belugacdn/grafana-belugacdn-app",
       "versions": [
         {
-          "version": "1.1.1",
-          "commit": "6cb4392e6877584b17642445bc392f9bb7117ddb",
+          "version": "1.2.0",
+          "commit": "ba2f5f8c1d285ee3fc2ae750a37f71d4a56a9e97",
           "url": "https://github.com/belugacdn/grafana-belugacdn-app"
         },
         {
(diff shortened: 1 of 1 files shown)


140am:
https://github.com/belugacdn/grafana-belugacdn-app/compare/v1.1.0...v1.2.0

daniellee:
Published and merged. The app gets a new url on Grafana.net: https://grafana.net/plugins/belugacdn-app

There are now two versions of the beluga app on grafana.net but one is hidden. Version 1.0.0 is not visible as the plugin.json contains the old plugin id.

This means two things:

1. If you want to have v1.0.0 available you will have to create a branch (branching from the 1.0.0 commit) and change the plugin id. I can then publish v1.0.0 with the new commit hash. It does not have to be in the master branch.

2. For users stuck with the old plugin id, you could do a version with a notice (a text panel on the dashboard for example) telling them how to uninstall and install the new plugin.

The old url still works (https://grafana.net/plugins/grafana-belugacdn-app) but is not in the plugin list any more.

Really sorry for this mess.

140am:
@daniellee  i don't see 1) being an issue as installations using the new plugin ID will always use 1.2+ and your idea for 2) is great to address existing installations. we will create a branch with re-installation instructions on the dashboard as you described and will reach out to you next week about the commit hash once that is ready.

thank you and have a great weekend!

PR #72, opened 2016-12-10
Monitoring Art datasource v1.0.1

jangaraj:
Optimized README for grafana.net (dist/README.md) + fixes for minor issues.
Is it possible to define "Monitoring Art" datasource type in dashboards section https://grafana.net/dashboards?

### DIFF
--- a/repo.json
+++ b/repo.json
@@ -895,15 +895,15 @@
     },
     {
      "id": "grafana-monitoring-art",
-      "type": "datasource",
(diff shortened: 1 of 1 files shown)


daniellee:
Published. Nice work with the README. That's an approach we should put in the documentation so that other projects can have two readmes if they want.

I kept it at version 1.0 as it was just a metadata change.

PR #75, opened 2016-12-12
add new panel type briangann-datatable-panel

briangann:
This is a new panel for Grafana based on Datatables.net. I've tested it with the default datasource and InfluxDB.

Please review and let me know if there are any changes needed.

Thanks!

Brian

### DIFF
--- a/repo.json
+++ b/repo.json
@@ -783,6 +783,18 @@
       ]
     },
     {
+      "id": "briangann-datatable-panel",
+      "type": "panel",
+      "url": "https://github.com/briangann/grafana-datatable-panel",
+      "versions": [
+        {
+          "version": "0.0.1",
+          "commit": "c4d3b04b52b1018ca000e4f34db6d008fc02c79c",
+          "url": "https://github.com/briangann/grafana-datatable-panel"
+        }
+      ]
+    },
+    {
       "id": "foursquare-clouderamanager-datasource",
       "type": "datasource",
       "url": "https://github.com/foursquare/datasource-plugin-clouderamanager",
--- a/repo.json
+++ b/repo.json
@@ -789,7 +789,7 @@
       "versions": [
         {
           "version": "0.0.1",
-          "commit": "c4d3b04b52b1018ca000e4f34db6d008fc02c79c",
+          "commit": "6b0ab7bdbfdecac403cc5b8bdf23f692be109c7a",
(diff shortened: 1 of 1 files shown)


daniellee:
Thanks for submitting this! Will get back to you with feedback soon.

daniellee:
Hi Brian,

I have started reviewing your plugin and I must say you have put an impressive amount of work into it!

It looks you have very ambitious plans for different themes. I'm curious about the themes you plan to support. How is that going to work in the future? Will you use the default theme from each framework or will the user be able to upload their own css?

### Feedback
The metadata and the readme look fine. Lots of screenshots which is great for panel plugin.

Some feedback on the plugin:

1. Functionality-wise the plugin works great - sorting, paging etc.
2. There are a few options that seem to require saving and then refreshing the page: striping, hover, cell borders, row borders. That wasn't immediately obvious to me - I thought they were just broken initially. Maybe a note in the popover to say a page refresh is needed? 
3.With Grafana's light theme everything is invisible (except for the paging text which is much clearer than on the dark theme):

image

If you could tweak the css for the light theme, then this is good to publish.

briangann:
For the themes:
  The default theme for each type is going to be used, then a page of coloring options that let you modify the different components. It's more to get a look&feel vs custom css.  I just have to sort out a workaround for the js import requirements.

I've updated tooltips for the options that require a page refresh.

I put on the patented Torkelo Grafana Light Theme Sunglasses and fixed the "light" theme :)

I'm going to adjust the dark theme so the buttons show up better, they're definitely too dark.

I'll have this cleaned up over the next day.

Thanks for the feedback!

briangann:
Hopefully I've squashed this PR correctly - if not let me know.

I found a couple of other bugs and have checked in those as well. New screenshots have been made, including the the light theme.

I've tested this with Influx with multiple panels, multiple rows, and multiple queries.  The theme option is still limited to just "Basic".

Alternative themes can work, it's all coded up but disabled. The side effect when enabled is that multiple datatable panels on the same dashboard cannot have different themes. More debugging is needed to make them independent. If it gets too unruly the theme option will be removed :)

Thanks!

daniellee:
Great work! Published and merged: https://grafana.net/plugins/briangann-datatable-panel

I look forward to see how this develops in the future. We have talked about improving the table panel so it's great to see someone from the Grafana community taking it on.

PR #81, opened 2016-12-22
rev plugin to v0.0.2 for latest bugfixes

briangann:
New version of plugin that addresses 4 bugs filed.

Let me know if there is anything else that needs to be added.

Thanks!


daniellee:
Merged and published.

PR #83, opened 2017-01-05
Add Satellogic 3D Globe Panel 0.1.0

peregrinogris:
This new plugin allows you to add a cesium panel to your dashboard, and define a custom URL for your CZML data file to add elements to the map.

Not sure if this is the right place, but this plugin should be published under https://grafana.net/orgs/satellogic in grafana.net

### DIFF
--- a/repo.json
+++ b/repo.json
@@ -958,6 +958,18 @@
           "url": "https://github.com/Vonage/Grafana_Status_panel"
         }
       ]
+    },
+    {
+      "id": "satellogic-3d-globe-panel",
+      "type": "panel",
+      "url": "https://github.com/satellogic/grafana-3d-globe-panel",
+      "versions": [
+        {
+          "version": "0.1.0",
+          "commit": "0930ebe33a023b71d0a55c9be29566b6fe8e231d",
+          "url": "https://github.com/satellogic/grafana-3d-globe-panel"
+        }
+      ]
     }
   ]
 }


daniellee:
This is the right place! Thanks for submitting this. I'll get back to you with feedback soon.

daniellee:
My feedback so far:

- The meta data (plugin.json) looks fine and the readme renders in grafana.net like it should. Maybe you could add a few lines on how to get up and running with a Cesium server? Something like a link to the [download page](http://cesiumjs.org/downloads.html) and maybe a link to the [CORS Chrome plugin](https://chrome.google.com/webstore/detail/allow-control-allow-origi/nlfbmbojpeacfghkpbjhddihlkkiljbi) if running the server on localhost. And test it with the included sample satellites data at localhost:5000/Apps/SampleData/simple.czml.

- The default is to render in the 2D view which was surprising for me as I was expecting 3D. Could the 3D view be set as default?

- The plugin doesn't save any settings except for url and bing api key. This means it does not remember which view I want 2D/3D or which map/terrain imagery I chose or zoom level.

- There doesn't seem to be any way to specifiy a starting point to center the map/globe? For example, if I always want the globe to be centered on Europe.

- A few of the map options in the cesium imagery menu returned authorization errors.

- The cesium plugin looks great and everything seemed to work fine when I loaded in a sample czml file.

- The built-in full screen button doesn't really do much. It does make the screen full-screen but the map is the same size.

There aren't any bugs that I could find but not being able to save a point to center the map and a start zoom level seems like a basic requirement to me. Or is that something you would do in the czml file?

peregrinogris:
Ok, let's address those comments (I know the PR has conflicts, but lets fix those once we reach a consensus over the comments you made :slightly_smiling_face:). One thing to keep in mind, is that this is the first release of this plugin. A `"version": "0.1.0"` is expected to have a few quirks, but we thought it's good enough to get the ball rolling and pusblish it to grafana.net.

>    The meta data (plugin.json) looks fine and the readme renders in grafana.net like it should. Maybe you could add a few lines on how to get up and running with a Cesium server? Something like a link to the download page and maybe a link to the CORS Chrome plugin if running the server on localhost. And test it with the included sample satellites data at localhost:5000/Apps/SampleData/simple.czml.

Those seem to be comments better addressed in Github's readme, not the Grafana's one. The way I see it, grafana.net readme should comment about the features of the plugin and how to use it, basically what you need to do after hitting "Install Plugin". Maybe adding a comment leading the potential user of the plugin to head to github for a more detailed description on how to locally test the plugin?

>    The default is to render in the 2D view which was surprising for me as I was expecting 3D. Could the 3D view be set as default?

The use case that triggered this plugin was for monitoring satellite positions along with other data in a grafana dashboard. One requirement was that the view were 2D by default, where you could see the whole orbit of the satellite in one look without having to turn the map. I can create an issue to make this a configuration, but it's an improvement.

>    The plugin doesn't save any settings except for url and bing api key. This means it does not remember which view I want 2D/3D or which map/terrain imagery I chose or zoom level.

Again, those features were not a usecase for this plugin, can be improvements for future versions though.

>    There doesn't seem to be any way to specifiy a starting point to center the map/globe? For example, if I always want the globe to be centered on Europe.

Another improvement for future versions.

>    A few of the map options in the cesium imagery menu returned authorization errors.
Mhh, this is a good one, didn't happen when we were testing it, can you elaborate a bit more so I can try reproducing them?

>    The built-in full screen button doesn't really do much. It does make the screen full-screen but the map is the same size.
I think that button is conflicting with grafana's fullscreen. It may be triggering the same event but grafana taking over it resulting in grafana being fullscreen and cesium not changing. Do you consider this a breaking issue?

Let me know what you think!

daniellee:
Don't worry about the merge conflict, I'll fix that.

None of my comments were show stoppers. Looking forward to future improvements then!

Published: https://grafana.net/plugins/satellogic-3d-globe-panel

PR #93, opened 2017-02-14
Add the InfluxDB admin panel

ryantxu:
Here is a simple panel that show current queries and visualizes raw query output

### DIFF
--- a/repo.json
+++ b/repo.json
@@ -915,13 +915,13 @@
       ]
     },
     {
-      "id": "natel-influx-admin",
+      "id": "natel-influx-admin-panel",
       "type": "panel",
       "url": "https://github.com/NatelEnergy/grafana-influx-admin",
       "versions": [
         {
           "version": "0.0.1",
-          "commit": "74299afadce9e4dffe77a860c59cdf1a21709694",
+          "commit": "49291c61630aedb01f5d5ae3284dba6c1bc9d332",
           "url": "https://github.com/NatelEnergy/grafana-influx-admin"
         }
       ]
--- a/repo.json
+++ b/repo.json
@@ -931,7 +931,7 @@
       "versions": [
         {
           "version": "0.0.1",
-          "commit": "49291c61630aedb01f5d5ae3284dba6c1bc9d332",
+          "commit": "6baca955aeca2156181065df7e4a838cc36b7a11",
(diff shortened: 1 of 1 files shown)


daniellee:
Can you change the plugin id as it is not valid right now -> http://docs.grafana.org/plugins/developing/code-styleguide/#metadata

A simple fix would be to add panel at the end, so it would become `natel-influx-admin-panel`

Thanks!

ryantxu:
I just changed the ID, let me know if you need anything else

daniellee:
The plugin.json file in the dist directory is the one that is used by Grafana, could you update the plugin id there as well. Thanks!

daniellee:
The panel looks great! Pity that we don't have anything like this for Graphite. Love the query window and how you can get all the Influxdb meta data.

For current queries, is it only long running queries that are visible? I tried opening up an Influx dashboard and nothing showed up in the panel.

There were a couple of things that weren't working so I sent a PR for them:

https://github.com/NatelEnergy/grafana-influx-admin/pull/1

ryantxu:
Thanks.

It shows all queries... But you are only likely to see ones that take more
than 1s

daniellee:
I messed up the PR, forgot to run grunt before committing (doh). I sent a new PR with a fix: https://github.com/NatelEnergy/grafana-influx-admin/pull/2

DanCech:
@ryantxu this is looking great, a couple more tweaks and it should be ready to go.  You'll want to set the updated date in plugin.json and I also noticed a typo in the name of the first screenshot.

Something else I'd recommend (though not required) is that moving plugin.json into src/ can simplify the image paths and gruntfile, you can see an example of that at https://github.com/grafana/simple-json-datasource

Finally, you'll want to log into your grafana.net account and register an org with the org-username "natel" which is what the plugin will be published under (the first segment of the plugin id/slug has to match the org-username).

ryantxu:
I just made updates.  Let me know if I'm missing anything

DanCech:
@ryantxu I just opened a PR in your repo with some tweaks

daniellee:
Published: https://grafana.net/plugins/natel-influx-admin-panel

Fantastic work with all three plugins @ryantxu I think you must be the person who has contributed most plugins!

PR #94, opened 2017-02-14
add clickhouse-grafana datsource

hagen1778:
Hi! We've created datasource plugin for [ClickHouse](https://github.com/yandex/ClickHouse) database

ps plugin should be published under https://grafana.net/orgs/vertamediallc in grafana.net

### DIFF
--- a/repo.json
+++ b/repo.json
@@ -1022,6 +1022,18 @@
           "url": "https://github.com/openstack/monasca-grafana-datasource"
         }
       ]
+    },
+    {
+      "id": "clickhouse-grafana",
+      "type": "datasource",
+      "url": "https://github.com/Vertamedia/clickhouse-grafana",
+      "versions": [
+        {
+          "version": "1.0.0",
+          "commit": "d55f7a36e40642be64879740e719409a3742488c",
+          "url": "https://github.com/Vertamedia/clickhouse-grafana"
+        }
+      ]
     }
   ]
 }
--- a/repo.json
+++ b/repo.json
@@ -1030,7 +1030,7 @@
       "versions": [
         {
           "version": "0.0.1",
-          "commit": "103264210c1547d4182c79f2dfadecb442d75fcf",
+          "commit": "7ec2bb0efa5972e283c9cbe5e76d930ead94a18b",
           "url": "https://github.com/Vertamedia/clickhouse-grafana"
         }
       ]
--- a/repo.json
+++ b/repo.json
@@ -1024,13 +1024,13 @@
       ]
     },
     {
-      "id": "clickhouse-grafana",
+      "id": "vertamedia-clickhouse-datasource",
       "type": "datasource",
       "url": "https://github.com/Vertamedia/clickhouse-grafana",
       "versions": [
         {
-          "version": "1.0.0",
-          "commit": "d55f7a36e40642be64879740e719409a3742488c",
+          "version": "0.0.1",
+          "commit": "103264210c1547d4182c79f2dfadecb442d75fcf",
           "url": "https://github.com/Vertamedia/clickhouse-grafana"
         }
       ]
--- a/repo.json
+++ b/repo.json
@@ -1030,7 +1030,7 @@
       "versions": [
         {
           "version": "0.0.1",
-          "commit": "7ec2bb0efa5972e283c9cbe5e76d930ead94a18b",
+          "commit": "1da0a28e68dfb5f8fd87d804865bd0db0e49fe46",
           "url": "https://github.com/Vertamedia/clickhouse-grafana"
         }
       ]


daniellee:
Thanks for submitting this plugin! Very cool to get a new data source. To get the plugin published under vertamediallc, you will need to change the plugin id to `vertamediallc-clickhouse-datasource`

Here are the [conventions](http://docs.grafana.org/plugins/developing/code-styleguide/#metadata) for the plugin id.

Any tips on how to set up Clickhouse locally? I'm thinking this [server](https://hub.docker.com/r/yandex/clickhouse-server/) on docker hub and then the [client](https://hub.docker.com/r/yandex/clickhouse-client/) to create some test data.

hagen1778:
> To get the plugin published under vertamediallc, you will need to change the plugin id to vertamediallc-clickhouse-datasource

The organization name `Vertamedia` was already taken at grafana.net by another organization. That's why i've added `llc` part. Originally github-account is https://github.com/Vertamedia
So this scheme `[github username/org]-[plugin name]-[datasource|app|panel]` confusing me. Would be plugin-id `vertamediallc-clickhouse-datasource` correct in this case?

> Any tips on how to set up Clickhouse locally? 
We used next instruction: 
- goto  https://clickhouse.yandex/
- click button `Download`
- follow instruction

Remember that CH supports date-columns [Date](https://clickhouse.yandex/reference_en.html#Date) and [DateTime](https://clickhouse.yandex/reference_en.html#DateTime). To build some fancy graphics you will need to create table with DateTime column. There is a little description about this columns in plugin [readme](https://github.com/Vertamedia/clickhouse-grafana#query-builder).

If you have any questions, I will be happy to answer them.

daniellee:
The scheme for plugin id is just a convention and assumes that Github org and Grafana.net org are the same. This is the first time we've had a conflict. I am planning to contact the current owner of the Vertamedia org and ask if they will change the org name, if that's ok with you? 

If they won't change the name, then the plugin id will have to be: `vertamediallc-clickhouse-datasource`.

Very nice work on the plugin! Here's my feedback:

- The query builder worked great for simple functions but it would be great if you could add an option there for group bys (or options for $columns, $rate, $rateColumns)

- Loved the macros, really simplified writing the sort of queries that most people would want to write. I have to admit that I got stuck for a while because I missed that the query has to start with the macro function i.e. I wrote `select $columns(Name, sum(Value) c)  FROM test.timeseries` instead of `$columns(Name, sum(Value) c)  FROM test.timeseries`. I saw it in the README later (`Query must begins from function name`) but maybe you could emphasize that more. Being able to see the expanded macros underneath is also a great feature.

- The group by time interval worked well. You have lots of help options that helped explain it too which is great.

Found one error:

https://github.com/Vertamedia/clickhouse-grafana/blob/master/clickhouse/query_part_editor.js#L17

The path to the partial should be `public/plugins/clickhouse/partials/query_part.html` instead.

Only two things that need to be fixed to get this published: the above error and getting the plugin id fixed. Everything else looks great. 

It was also very interesting for me to get a first look at Clickhouse. It looks impressive. Are you doing rollups or just using raw data? I noticed that there is a [graphite cluster](https://github.com/lomik/graphite-clickhouse) project that uses [graphite rollups in Clickhouse](https://github.com/yandex/ClickHouse/blob/master/dbms/include/DB/DataStreams/GraphiteRollupSortedBlockInputStream.h).

P.S. Don't worry about the merge conflict, I'll fix that.

hagen1778:
Thanks for feedback!
> I am planning to contact the current owner of the Vertamedia org and ask if they will change the org name, if that's ok with you?
Yes, it's ok. I will wait for your answer about correct plugin id

> maybe you could emphasize that more
I moved functions description to the beginning of paragraph. Hope this helps

> Found one error
I've fixed it. It was connected with way how we using Grafana, where backend and frontend parts of plugin were placed directly in internal folders.

> It looks impressive
Yes! We have a really huge amount of data and CH helps us to handle it.

> Are you doing rollups or just using raw data?
In most cases we use raw data. But it depends on a task. Sometimes we use SummingMergeTree or AggregatingMergeTree engines to aggregate data before output it to dashboard.

daniellee:
Good news, you now have the Vertamedia org: https://grafana.net/orgs/vertamedia The person who had the name had no problem changing it.

If you can add a plugin.json file then we are good to go. 

http://docs.grafana.org/plugins/developing/plugin.json/

hagen1778:
Thanks! I've changed [plugin.json](https://github.com/Vertamedia/clickhouse-grafana/commit/103264210c1547d4182c79f2dfadecb442d75fcf). I hope all is well

daniellee:
One last thing, the path is wrong here:

https://github.com/Vertamedia/clickhouse-grafana/blob/master/clickhouse/query_part_editor.js#L17

should be `public/plugins/vertamedia-clickhouse-datasource/partials/query_part.html`

hagen1778:
Yes, totally forgot about this. Fixed now
Should i rm `clickhouse` folder and move all files to the root?

daniellee:
Sorry about the slow reply was out of town the last two days. Yes, it would be great if you could rm the clickhouse folder and move it all to the root. Thanks!

hagen1778:
Done!

daniellee:
Finally, thanks for being so patient!

Published: https://grafana.net/plugins/vertamedia-clickhouse-datasource

hagen1778:
Thanks!

PR #97, opened 2017-02-17
add plotly panel (scatter plots)

ryantxu:


daniellee:
This is a very nice first version. What are your plans for this? Are you going to concentrate on Scatter plots or expand it to other plotly graphs as well?

Published: https://grafana.net/plugins/natel-plotly-panel

ryantxu:
Thanks!  our current focus in on the 2D scatter plot.  The near term work is making the merging data from various sources better.  (Right now it gives errors unless the samples are identical) When you graph data that is sampled at 10hz with data at 1hz, it could just pick the points that match.

Plot.ly has LOTS of nice features and options, so i hope others can explore them for their needs.

PR #98, opened 2017-02-17
add discrete panelel

ryantxu:


daniellee:
Great panel! Seems to be bug free and it is a very nice visualization.

Published: https://grafana.net/plugins/natel-discrete-panel

PR #104, opened 2017-03-04
Update D3Gauge plugin version to 0.0.3

briangann:
This release fixes a number of bugs and adds some new features.

Bugfixes:
 * Extra gauge no longer appears when using "duplicate"
 * Now correctly sizes according to panel settings (Height and Width)
 * Needle animation works as intended 

New Features:
  * Needle animation speed is now configurable
  * Arbitrary degree gauges now supported (default is from 60 to 320)
  * Start of 0 value in gauge can be set to any degree
  * Value text on gauge can now be moved up/down as needed (Useful for full-circle gauges)

### DIFF
--- a/repo.json
+++ b/repo.json
@@ -826,6 +826,11 @@
       "url": "https://github.com/briangann/grafana-gauge-panel",
       "versions": [
         {
+          "version": "0.0.3",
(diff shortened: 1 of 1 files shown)


daniellee:
Wrong SHA but otherwise looks great! Published and I'll fix it so that the right SHA is merged here.

PR #109, opened 2017-03-23
Add Digiapulssi Organisations Panel

timyllym:
This panel displays all organisations that current user belongs to as navigable links (buttons).

### DIFF
--- a/repo.json
+++ b/repo.json
@@ -1188,6 +1188,18 @@
           "url": "https://github.com/sni/grafana-pnp-datasource"
         }
       ]
+    },
+    {
+      "id": "digiapulssi-organisations-panel",
+      "type": "panel",
+      "url": "https://github.com/digiapulssi/grafana-organisations-panel",
+      "versions": [
+        {
+          "version": "1.0.0",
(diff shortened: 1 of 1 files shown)


daniellee:
Great to see that you are contributing another new plugin!

Could you fix the plugin id: https://github.com/digiapulssi/grafana-organisations-panel/blob/master/dist/plugin.json#L4 to follow the [plugin id conventions](http://docs.grafana.org/plugins/developing/code-styleguide/#plugin-json-mandatory) and then update the PR with the new commit SHA. Thanks! (And remember it is the plugin.json in the dist directory that Grafana tries to read so that is the important file to update with the correct plugin id).

I think that `digiapulssi-organisations-panel` would be a suitable plugin id.

timyllym:
Sure. I forgot to run grunt. it's now fixed.

daniellee:
Merged and published: https://grafana.com/plugins/digiapulssi-organisations-panel

timyllym:
Thanks! We have hidden the top bar from our users so there needs to be another way to navigate through organisations. We have custom home page that shows the organisations.

organisations-screenshot-light

daniellee:
Nice! I'd recommend adding that screenshot to the readme.

PR #112, opened 2017-04-04
Added annotations-plugin

naiconovalabs:


DanCech:
This looks good at first glance 👍 

@naiconovalabs you may want to move `plugin.json` into `src/` to avoid any confusion with the image paths, once you do that you can also remove the special handling for it in the grunt file.

daniellee:
@naiconovalabs Thanks for sending this in. I believe this is one of the most requested features for Grafana. Here is my feedback so far.

## Metadata

There are two broken image links in the readme. They work on GitHub because GitHub fixes links for internal links. To work on grafana.com and GitHub, the links should be:

- https://raw.githubusercontent.com/novalabs/grafana-annotations-panel/master/src/img/screenshot1.png

- https://raw.githubusercontent.com/novalabs/grafana-annotations-panel/master/src/img/screenshot2.png

The instructions in the readme need some improvements:

- Explain that this is an annotations data source with a link to the Grafana docs -> http://docs.grafana.org/reference/annotations/#influxdb-annotations
- Explain that the annotation will be matched by the tag key.

For bonus points, a gif or gifs showing how to create an annotation would be fantastic.

## Code

- The default data source is [hard coded to be lowercase `influxdb`](https://github.com/novalabs/grafana-annotations-panel/blob/master/src/annotations_ctrl.js#L16-L17). This approach will not work for most people. It is quite common to have multiple InfluxDB data sources or in my case, my data source was named `InfluxDB`. You need to find all data sources of type `InfluxDB` using the `backendSrv` service instead.
- When you click the `Save Annotation` button you do not get any feedback if it has succeeded or not. The `alertSrv` service can be used for this. Something like this: `this.alertSrv.set('Saved', 'Successfully saved the annotation', 'success', 3000);`
- The options tab should be named Annotation Options (remove the underscore).

An example taken from another plugin of fetching data sources by type:
```js
  getGraphiteDatasources() {
    var self = this;
    return this.backendSrv.get('/api/datasources')
    .then((result) => {
      self.datasources = _.filter(result, {"type": "graphite"});
    });
  }
```

Once I renamed my data source to lowercase influxdb, then the annotations worked great. Let me know if you need any help.

naiconovalabs:
Hi.
Thanks for suggestions.
Here my fixes:

- moved `plugin.json` in `/src`
- fixed images link in README
- added a gif image
- added Success/Error popup (using alertSrv)
- added datasources list

@daniellee It's not clear to me what you mean with "Annotation will be matched by the tag key". Users can insert custom query, matching not only tags but also title or text or nothing...

There is some documentation for `alertSrv`, `backendSrv` or `somethingelseSrv` ?

Thanks so much

daniellee:
Merged and published: https://grafana.com/plugins/novalabs-annotations-panel

I apologize for the delay. Been busy with the 4.3.0 release. Thanks for you patience.

Since you submitted this, we started on working on adding [better annotation support](https://github.com/grafana/grafana/issues/1286) to Grafana. It should be released in v4.4.0. It overlaps with your plugin but the major difference is that the data is saved to the Grafana data source (mysql, sqlite or postgres) rather than to InfluxDB. So I think your plugin will be good option for people who want to save the annotation data in their TSDB rather than in Grafana.

There is unfortunately no documentation for alertSrv but there are lots of examples. (I will add some documentation on these to the docs site).

To use alertSrv 
- first you inject into your constructor like [this](https://github.com/grafana/kubernetes-app/blob/master/src/components/clusters/podInfo.js#L10)
- then you call it like [this](https://github.com/grafana/kubernetes-app/blob/master/src/components/clusters/podInfo.js#L20)

It takes 3 mandatory parameters, title, body and severity plus one optional parameter - timeout. Severity can be `error`, `warning` or `info`. Timeout is used to fade out the popup after a specified time. 

Thanks for your patience! Any updates you want to do I should be able to publish them very quickly. Just remember to update the version number in the plugin.json file in your project and then submit the commit SHA and version number in a PR here.

PR #119, opened 2017-05-03
Add DalmatinerDB datasource

Licenser:


daniellee:
Very exciting! We were just talking about DalmatinerDB the other day. Great to get it into Grafana. 

I'll get back to you with feedback in a day or two.

Licenser:
:D take your time, I'm sure there is quite a bit to comment on and improve ;)

daniellee:
Usually I have tons of feedback for a new data source plugin but this was very good for a first version. Once I found the docker-compose repo for DalmatinerDB, it was easy to get everything working. It is definitely ready to publish.

There is one small thing to fix in the meta data - the plugin id. In this PR, the id is projectfifo-dalmatinerdb-datasource and in the plugin.json, the plugin id is dalmatinerdb-datasource. Which one do you want? Both follow the convention and are valid (as you are also the owner of the dalmatinerdb GitHub org).

Other feedback:

- Love the query editor. Looks and works great.
- I was not sure about how to use the Alias field though. You could add an editor help section like the Graphite and InfluxDB data sources have. Here is how the [InfluxDB data source](https://github.com/grafana/grafana/blob/master/public/app/plugins/datasource/influxdb/partials/query.options.html#L16-L21) does it (and here is the [text for the editor help](https://github.com/grafana/grafana/blob/master/public/app/plugins/datasource/influxdb/partials/query.options.html#L42-L52)).
    
image

- No problems with the add data source page either. It tests that the data source works which is the important thing.
- The README was ok but a bit out of date. It refers to Grafana 3 and has installation instructions that include cloning the repo. So it could do with some more work but it is not a showstopper.
- No logo. Do you need help with that? I could ask our Graphics guy if he has time to design one.

Licenser:
Glad that you like it. For the record all of the credit regarding the quality goes to @szarsti and @davecromberge they're quite awesome with this things :D

I'll update the plugin ID to `dalmatinerdb-datasource` shorter is better I looked at other plugins and had the impression the 3 section one is the default if two are OK.

The explanation of the alias is a very good point, some info is good I'll add that :) and do a one over on the readme while I'm at it. :)

daniellee:
@Licenser do you want me to publish this now or wait until you make some changes?

Licenser:
Sorry for the silence, I got a bit carried away with other things, I've nearly everything done just was stuck on adding a help link for the naming and ended up in a javascript rabbit-hole and drank enough coffee to forget what I wanted to do - I was happy for a moment! I'll see if I can figure it out tomorrow :)

Licenser:
Okay all sorted :D

daniellee:
One tiny change needed. The plugin.json file in the dist folder has version 1.0.0 and not 1.0.3. As this plugin has not been published yet, you can change it back to 1.0.0 in the plugin.json in your src folder if you prefer.

Licenser:
I can't bring myself to change tags once I pushed them ;) so it's `1.0.4` now but it's consistent over all places :D

Licenser:
a slightly unrelated question, is it possible to use a tag instead of a hash?

daniellee:
Merged and published: https://grafana.com/plugins/dalmatinerdb-datasource

Thanks so much for taking the time to submit your plugin!

The version needs to be connected to a specific hash and a tag could potentially point to different hashes (if you force push). But practically a tag or hash doesn't really matter. I can get a commit hash from a tag.

PR #122, opened 2017-05-12
Update vertamedia-clickhouse-datasource

hagen1778:
Hi! We've updated plugin with additional examples in README, syntax highlighting and other

### DIFF
--- a/repo.json
+++ b/repo.json
@@ -1204,8 +1204,13 @@
       "url": "https://github.com/Vertamedia/clickhouse-grafana",
       "versions": [
         {
+          "version": "0.0.1",
+          "commit": "1da0a28e68dfb5f8fd87d804865bd0db0e49fe46",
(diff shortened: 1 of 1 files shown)


daniellee:
Two things to fix:

- Update the [version in the plugin.json](https://github.com/Vertamedia/clickhouse-grafana/blob/master/plugin.json#L27) to be 0.0.2 so that it matches the version number in this PR. 
- In the PR, rather than replacing 0.0.1, add 0.0.2 (like [this](https://github.com/grafana/grafana-plugin-repository/blob/master/repo.json#L7-L18)) to the version list.

Thanks!

hagen1778:
Thanks!

daniellee:
Published and merged. Love the changes in the readme. (I really like the query editor for Clickhouse but I think I said that before).

hagen1778:
@daniellee Thank you for the kind words!

PR #124, opened 2017-05-18
Update btplc-alarm-box-panel

benthorner:
Add Values tab to btplc-alarm-box-panel to allow advanced users to change the way the display and color values are calculated.


daniellee:
Sorry for the delay. Got an error when generating a thumbnail from one of the images and haven't had time to investigate it yet - not sure if the problem is with the image or with our code :grimacing:

daniellee:
It is a bug in the grafana.com api. I have sent a fix so I should be able to publish this later today.

daniellee:
Merged and published!

benthorner:
Thanks Dan - that's great!

PR #126, opened 2017-05-25
added savantly-heatmap-panel v0.2.0

jdbranham:


DanCech:
Published at https://grafana.com/plugins/savantly-heatmap-panel

DanCech:
@jdbranham Scratch that, plugin.json version needs to be updated

PR #127, opened 2017-05-29
Update to new dalmatiner plugin version

Licenser:
1.0.4 of the DalmatinerDB plugin had a bug in calculating times for the graph, 1.0.5 fixes this


daniellee:
Merged and published.

PR #128, opened 2017-05-30
Add Gnocchi datasource

sileht:


sileht:
Can we merge this ?

daniellee:
Sorry for the delay. I haven't had time to fix a redirect in Grafana.com yet so that the plugin id can be changed without breaking updates for all the current users. We have support for deprecating the previous plugin id but not for redirects yet.

sileht:
Maybe we can merge that and have the redirect later ?

daniellee:
If I do that there will be two Gnocchi data sources. I could deprecate the old one. This means users will be informed that it is deprecated in Grafana and it won't show up on Grafana.com.

sileht:
Sorry, @pastamaker-bot was me, I was logged with a wrong account.

daniellee:
New data source published here: https://grafana.com/plugins/gnocchixyz-gnocchi-datasource

And the old data source is deprecated: https://grafana.com/plugins/sileht-gnocchi-datasource
which means it will not be visible on the plugins page on grafana.com anymore.

sileht:
Great, thank a lot !

PR #129, opened 2017-06-01
Update repo.json

jdbranham:
Bumped revision for Savantly Heatmap

### DIFF
--- a/repo.json
+++ b/repo.json
@@ -820,7 +820,7 @@
         },
         {
           "version": "0.2.0",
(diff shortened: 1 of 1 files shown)


daniellee:
Hi Jeremy, can you bump the version to 0.2.0 [here](https://github.com/savantly-net/grafana-heatmap/blob/master/dist/plugin.json#L17) otherwise users won't see that there is a new update in Grafana (as they already have version 0.1.0). Thanks.

daniellee:
Merged and published

PR #130, opened 2017-06-02
Logstreaming Plugin

Chris7:
This adds in a plugin that can stream data from a data source into a panel. It is useful for log aggregation across multiple services.

### DIFF
--- a/repo.json
+++ b/repo.json
@@ -1307,7 +1307,7 @@
       "versions": [
         {
           "version": "0.0.1",
-          "commit": "83420b256e5aa39807dffdebeef132a37b75a1f5",
+          "commit": "8455536a1712b0887a2ab1815bb6716eb29529d1",
           "url": "https://github.com/ginkgobioworks/grafana-logstream"
         }
       ]
--- a/repo.json
+++ b/repo.json
@@ -1301,6 +1301,18 @@
       ]
     },
     {
+      "id": "ginkgobioworks-grafana-logstream",
+      "type": "panel",
+      "url": "https://github.com/ginkgobioworks/grafana-logstream",
+      "versions": [
+        {
+          "version": "0.0.1",
+          "commit": "09a8f4448326dc89501c7ef5664d94abbeaf92ae",
+          "url": "https://github.com/ginkgobioworks/grafana-logstream"
+        }
+      ]
+    },
+    {
       "id": "adremsoft-netcrunch-app",
       "type": "app",
       "url": "https://github.com/adremsoft/grafana-netcrunch-plugin",
--- a/repo.json
+++ b/repo.json
@@ -1301,13 +1301,13 @@
       ]
     },
     {
-      "id": "ginkgobioworks-grafana-logstream",
+      "id": "ginkgobioworks-logstream-panel",
       "type": "panel",
       "url": "https://github.com/ginkgobioworks/grafana-logstream",
       "versions": [
         {
           "version": "0.0.1",
-          "commit": "09a8f4448326dc89501c7ef5664d94abbeaf92ae",
+          "commit": "83420b256e5aa39807dffdebeef132a37b75a1f5",
           "url": "https://github.com/ginkgobioworks/grafana-logstream"
         }
       ]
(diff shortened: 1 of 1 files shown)


daniellee:
Sorry for the delay here. I will be back home after travelling next week and will get back to you with feedback then. This looks very interesting!

daniellee:
Noticed that the plugin id is not following the [plugin id convention](http://docs.grafana.org/plugins/developing/code-styleguide/#plugin-json-mandatory).

`ginkgobioworks-grafana-logstream` should be `ginkgobioworks-logstream-panel` in the plugin.json file.

Chris7:
Thanks for the check, fixed the id.

daniellee:
@Chris7 could you explain how to use this plugin? To get streaming data does it rely on the auto refresh in Grafana? 

I tried with autorefresh every 5 seconds but I couldn't see that toggling the tail function did anything. Think I'm missing something here :-)

Chris7:
It should populate whatever is present from the query time. If you have a datasource like elasticsearch, use the raw document metric, then choose the columns to display in options. It works very much like the table display.

daniellee:
aha, got totally tricked by the name and was searching the code for streaming functionality (like the Snap plugin has). Will get back to you with feedback soon. Thanks for your patience!

daniellee:
Here's my feedback so far:

### Meta Data

- This [image link](https://github.com/ginkgobioworks/grafana-logstream/blob/master/dist/plugin.json#L15) in the plugin.json is incorrect. The link should be relative to the plugin.json file in the dist directory. One way to solve this is to create a src directory in the dist directory - an example [here](https://github.com/grafana/worldmap-panel/tree/master/dist/src/images).

- The plugin.json should contain a link to the GitHub project.

- The README.md should explain a bit more what the plugin does. (Just a little intro paragraph at the start - explain that it is an alternative to the table panel).

### Plugin

- The Sanitize HTML option in the column styles for strings does not work.

So just a few small changes and we're good to go! Thanks again for your patience.

daniellee:
Followup on the image link. It is easier to just move the plugin.json into the src directory instead of changing the link.

Chris7:
Hi @daniellee,

With respect to point 2, does https://github.com/ginkgobioworks/grafana-logstream/blob/master/dist/plugin.json#L18 not do that?

daniellee:
Aha, there is no plugin.json in the dist directory. That's why there was no link. Grafana loads the plugin from the dist subdirectory if it exists so everything the plugin needs must be in the dist directory.

Chris7:
Yes, I moved it per your suggestion :). I'll rebundle it

daniellee:
Sorry for the confusion here. The idea with a src and dist directory is that src is for development files and dist is for the minified/compiled/built files. The build step typically uses the src files to build the dist files (either by copying or generating a new compiled version). Everything the plugin needs should be in the dist directory. If there is a dist directory then that is the code that Grafana will attempt to load. It will ignore the src directory.

Right now the src and dist directories have some differences. The plugin.json file and img subdirectory are not in the dist directory. Also, the libs subdirectory is not in the src directory.

Chris7:
I updated the dist directory with a new release. By the libs subdir, do you mean bower_components/node_modules? They aren't related to the project and shouldn't be in the src directory. Grunt takes care of bundling whatever code is necessary and puts it all in dist.

daniellee:
Thanks for your patience! Could you fix the incorrect link to the image in the plugin.json file here: https://github.com/ginkgobioworks/grafana-logstream/blob/master/dist/plugin.json#L15

daniellee:
@Chris7 did you ever get time to fix that last error?

Chris7:
@daniellee 

I believe it is fixed, the image exists at the path (relative to `dist`).

Chris7:
I just updated the commit hash in the PR.

daniellee:
Sorry about that - I totally missed that you had made an update.

daniellee:
I apologize for the long delay on giving feedback. This panel does not work with Grafana 5.x as the dashboard grid now uses a different component so the height calculations in the plugin will not work anymore.

Are you still interested in getting this published?

daniellee:
Closing this - let me know if you want to reopen.

PR #134, opened 2017-06-19
Add Prometheus AlertManager datasource

cryptobioz:


daniellee:
Thanks for submitting this. Looks like it will be really useful.

Some feedback on the meta data needed to publish a plugin.

- Needs a README.md (with an explanation of what the data source is and some simple instructions on how to use it)
- The plugin id in the plugin.json does not follow the convention. `grafana-prometheus-alertmanager-datasource` should be `camptocamp-prometheus-alertmanager-datasource`

I'll get back to you with more feedback in the next few days.

cryptobioz:
Thanks for the feedback and sorry for the delay.
I changed the plugin Id and added a basic README.md.

daniellee:
Thanks for the update. Here's my feedback so far:

### Metadata

- The plugin id in the plugin.json file in the dist subdirectory is still incorrect: https://github.com/camptocamp/grafana-prometheus-alertmanager-datasource/blob/master/dist/plugin.json#L3
- the links to the images in the README.md need to be changed from relative urls to full GitHub urls. Like this: `https://raw.githubusercontent.com/camptocamp/grafana-prometheus-alertmanager-datasource/master/images/table.png`

### Plugin 

- The severity column in table mode expects a number but I can't find anything in the Prometheus or AlertManager docs that defines what severity should be. Most examples of severity values are strings like "critical" or "page". I'm not a prometheus expert but it seems like it should be able handle strings as well.
- The timeseries and single mode seem to be the same? Timeseries in this case seems sort of pointless. Maybe is should be removed or are you planning to do something more with it in the future?
- Would it be possible/useful to have columns for annotations in table mode?

Any plans to take this a bit further? Maybe an app that includes this data source or a panel that can use this data source would allow linking to the alertmanager instance - so that you could silence an alert or drilldown to a the query (like the source link on an alert in AlertManager).

This is pretty close to being mergeable/publishable. 

Hopefully [this](https://github.com/grafana/grafana/issues/7481) will be implemented in the near future as well. It would make a great combo with this plugin.

cryptobioz:
I changed the plugin id and the images links.

About the plugin :
> - The severity column in table mode expects a number but I can't find anything in the Prometheus or AlertManager docs that defines what severity should be. Most examples of severity values are strings like "critical" or "page". I'm not a prometheus expert but it seems like it should be able handle strings as well.

That's true, we manually set a severity number into our alerting rules but this could also be a string. I'll change this.

> - The timeseries and single mode seem to be the same? Timeseries in this case seems sort of pointless. Maybe is should be removed or are you planning to do something more with it in the future?

I'll check that but the timeseries format will probably be removed.

> - Would it be possible/useful to have columns for annotations in table mode?

That's a good idea, I'll try to add this.

> Any plans to take this a bit further?

I'm working on a panel with features that you describe but I'm not a Grafana expert so it takes time.

daniellee:
Really have to get some linting in place for the plugin.json file :pensive: 

When Grafana loads the plugin, it will only load the dist directory if it exists. This means in your case, it will not load the images subdirectory and therefore the image links will be broken. Can you move the images directory into your src and dist directory. You will also need to copy the README.md to the dist directory.

cryptobioz:
I did some updates of the plugin.json file and fixed the problem with the images and the README.

daniellee:
Thanks for your patience - finally merged and published: https://grafana.com/plugins/camptocamp-prometheus-alertmanager-datasource

Do you want me to ask our designer to make a logo for the plugin?

cryptobioz:
Thanks to you!
If your designer can make a logo, that would be great!

PR #135, opened 2017-06-19
Add status dot panel and update alarm box

benthorner:
We'd like to add a new Status Dot panel, which is similar to existing Trend Dot plugin, but less restrictive. It displays a dot for each (Graphite) series, colored using configurable expressions.

In fact, we'd like to discontinue/remove the btplc-trend-dot-panel plugin we published previously, as we don't use it and are unlikely to maintain it. It's pretty niche! What are your thoughts on this?

### DIFF
--- a/repo.json
+++ b/repo.json
@@ -35,6 +35,18 @@
       ]
     },
     {
+      "id": "btplc-status-dot-panel",
+      "type": "panel",
+      "url": "https://github.com/BTplc/grafana-status-dot",
+      "versions": [
+        {
+          "version": "0.0.1",
+          "commit": "c493f4924a5649cee42a2015231d2689ea73943a",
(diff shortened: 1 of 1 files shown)


daniellee:
It all looks good here. With regard to the trend dot plugin, it is totally up to you. I can mark it deprecated so that no new users will find it.

New plugin published: https://grafana.com/plugins/btplc-status-dot-panel

And update to alarm box published too.

benthorner:
Thanks Dan :-)

Please can you mark the Trend Dot plugin as deprecated.

daniellee:
Done. This means it is taken out of the list of plugins on grafana.com and marked as deprecated if you happen to go to the direct link: https://grafana.com/plugins/btplc-trend-dot-panel

benthorner:
Great, thanks Dan!

PR #140, opened 2017-07-04
Added HTML panel plugin

linar-jether:
a fork of the ajax plugin with support for iframes and template variables (Easily add panels from other systems)

### DIFF
--- a/repo.json
+++ b/repo.json
@@ -1399,7 +1399,7 @@
       "versions": [
         {
           "version": "0.0.1",
-          "commit": "44a83d83ddb5e276e6727958e1d3a9b8fa36aced",
+          "commit": "3f5cbd6a6388c195419a0379cf2ec2d4104ff066",
           "url": "https://github.com/jether-energy/grafana-html-panel"
         }
       ]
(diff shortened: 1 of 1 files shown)


linar-jether:
Added some fixes and updated the commit #

daniellee:
Thanks for sending this. The plugin looks fine both the metadata and your changes all work they way should (template variables and iframes). 

My only problem/question is: it is so similar to the Ajax panel plugin, could you not ask Ryan to merge these changes into his plugin?

If that is not possible for whatever reason then I can absolutely publish this.

linar-jether:
No problem with merging back to Ryan's repo, created a fork mainly because of it not being only ajax and more general purpose way to display panels/content from external systems (i thought about making this more like the text/raw html panel but with passing time range as arguments)

Think it would be better to merge back to the ajax panel? 

Either way is fine by me

daniellee:
Closing this as it looks like you are going to merge this into the Ajax panel instead.

Ref: https://github.com/jether-energy/grafana-html-panel/issues/1

gesaleh:
is the merge done ?
i don't see any iframe option in ajax plugin 
and another question how you pass the parameter from,to height in my case they are not even passed to iframe

linar-jether:
@gesaleh waiting for @ryantxu to do the merge

PR #142, opened 2017-07-07
Added Bubble-chart panel plugin.

digrich:
This panel is meant to have a centralized view of any component in the form of a bubble chart.

### DIFF
--- a/repo.json
+++ b/repo.json
@@ -1398,7 +1398,7 @@
       ]
     },
     {
-      "id": "bubblechart-panel",
+      "id": "digrich-bubblechart-panel",
       "type": "panel",
       "url": "https://github.com/digrich/bubblechart-panel",
       "versions": [
--- a/repo.json
+++ b/repo.json
@@ -1396,6 +1396,18 @@
           "url": "https://github.com/petrslavotinek/grafana-carpetplot"
         }
       ]
+    },
+    {
+      "id": "bubblechart-panel",
+      "type": "panel",
+      "url": "https://github.com/digrich/bubblechart-panel",
+      "versions": [
+        {
+          "version": "0.0.1",
+          "commit": "4882d57ab74b1eec94cddcbaeeef12e2cdcc47fd",
+          "url": "https://github.com/digrich/bubblechart-panel"
+        }
+      ]
     }
   ]
 }
--- a/repo.json
+++ b/repo.json
@@ -1404,7 +1404,7 @@
       "versions": [
         {
           "version": "0.0.1",
-          "commit": "663a7854d8d270fd0625ed4581bd3c6798481fdf",
+          "commit": "b8194655f84ab3a3528244fdcebd75f7c86cab4e",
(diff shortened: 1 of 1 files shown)


DanCech:
Thanks for the contribution!

The first thing I see is that the plugin id will need to be updated to be in the correct format for publishing on grafana.com, it needs to be `<your grafana.com username or org>-bubblechart-panel`.

You'll also need to update the image URLs in your README.md file to be absolute (eg `https://github.com/digrich/bubblechart-panel/blob/master/src/img/BC_O1.png`) so that they can be resolved properly by grafana.

Once those are taken care of we'll review in more detail ASAP.

digrich:
Thanks @DanCech. Made all changes.

daniellee:
There are a couple of more things to fix:

- We don't support HTML in the markdown of README.md. Can you strip this out and just have normal image links. This will probably mess up your layout a bit but you can resize the images (using imageMagick or some other program) rather than using css to control the size.

- The [path to the editor html file](https://github.com/digrich/bubblechart-panel) does not work. The path should be `public/plugins/digrich-bubblechart-panel/editor.html`.

- The labels and fields on the editor tab should be lined up. You can use the `ẁidth-x` CSS classes to control the width. For example, in the first column all the labels should have the same width as the Group Separator label and all the fields should be the same width as the unit field.
     
    image

- Maybe the default height for a Bubble Chart panel should be larger than the default 100px. The initial chart is quite small, maybe 300 or 400px would be better?

But otherwise a very cool plugin. Love the zoom in/zoom out effect when clicking on bubbles and then outside the bubble.

digrich:
Thanks, @daniellee for your review. I have incorporated all your feedback.

daniellee:
That was fast!

Merged and published: https://grafana.com/plugins/digrich-bubblechart-panel

It looks visually amazing. Just curious but how are you using this in production? And is this a special type of bubble chart? I think the only bubble chart I've seen before is the one in [Microsoft Excel](https://support.office.com/en-US/article/Present-your-data-in-a-bubble-chart-424D7BDA-93E8-4983-9B51-C766F3E330D9?ui=en-US&rs=en-US&ad=US&fromAR=1).

Anyway, great work!

daniellee:
One more small thing. I changed the version to 1.0.0 here in the repo.json to match the version that you had in the plugin.json.

digrich:
Thanks @daniellee. 

It's a d3 based bubble chart to represent the component hierarchy and status at the same time. Inspired by 'Mike Bostock' example 
https://bl.ocks.org/mbostock/4063530

In production, we have multiple instances of different types of applications and this chart helps to group them together to provide a global view. Just by seeing the chart on a big screen, we get a quick sense of applications health and also it tells if the issue is happening to only some specific type of applications or spreading across the applications. Also, it reflects what type of applications are consuming more system resources.

daniellee:
So it's like have lots of single stats in one panel? Pretty cool.

PR #143, opened 2017-07-09
Add a new app: GLPI

ddurieux:
Add a new app called 'glpi_app_grafana'. The goal is to get data in GLPI (asset management / servicedesk) and display them into Grafana ;)

### DIFF
--- a/repo.json
+++ b/repo.json
@@ -1408,6 +1408,17 @@
           "url": "https://github.com/ayoungprogrammer/grafana-finance"
         }
       ]
-    }
+    },
+    {
+      "id": "glpi_app_grafana",
+      "type": "app",
+      "url": "https://github.com/ddurieux/glpi_app_grafana/",
+      "versions": [
+        {
+          "version": "1.0.0",
+          "commit": "041ee2c50767d434c8a78f4b2e5c372b1b4d6c2a",
+          "url": "https://github.com/ddurieux/glpi_app_grafana/"
+        },
+     }    
   ]
 }


daniellee:
Thanks for submitting this. I just started looking at this. Most of the meta data looks good (especially the readme!) but I noticed that the plugin id in the plugin.json file does not follow the [convention](http://docs.grafana.org/plugins/developing/code-styleguide/#plugin-json-mandatory). It should be `ddurieux-glpi-app`.

I got GLPI up and running locally but got a bit stuck on the permissions for the rest api. I will try again tomorrow and get back to you with more feedback then.

ddurieux:
Oh sorry, I missed the id convention :(
I will fix it in few hours

daniellee:
I've gotten totally stuck on connecting to the api via Grafana. Have it working via postman but when I try to do in proxy mode in Grafana I keep getting IP filter errors saying IP address 127.0.0.1 is not allowed. Do you use proxy or direct? I can probably get it working with direct and changing the CORS headers in Apache but I'm wondering if that is the normal use case?

These are my api client settings (have tried a few different settings and ranges here):

image

daniellee:
Nevermind, got it working with CORS headers.

daniellee:
Thanks for your patience, here is my feedback:

### Metadata
- The readme is pretty good but maybe could include a few more details about how to enable the rest api and links to how to set CORS headers for apache/nginx.

### Plugin

For the query tab, is there anyway to abstract away the url? Starting here with "undefined", I'm not sure how I would build a query without the included dashboard:

 image

When I tried to copy and paste in queries it only worked sometimes. I don't have any good ideas on how to do it better as there seem to be a million possible parameters to include in a query. But maybe a few predefined common queries that the user can select from a dropdown would be a good start.

None of these are showstoppers and can be done in the next version.

Published: https://grafana.com/plugins/ddurieux-glpi-app

PR #55, opened 2016-11-02
Add hawkular datasource

jotak:

PR #146, opened 2017-07-18
Adding Sidewinder datasource

ambud:
Requesting to add Sidewinder TSDB's Grafana plugin to the Plugin List:
https://github.com/srotya/sidewinder

### DIFF
--- a/repo.json
+++ b/repo.json
@@ -351,6 +351,18 @@
       ]
     },
     {
+      "id": "sidewinder-datasource",
+      "type": "datasource",
+      "url": "https://github.com/srotya/sidewinder-grafana/",
+      "versions": [
+        {
+          "version": "0.0.1",
+          "commit": "0edb6374cc3ab128b6b9994a9f76a8c32109fd95",
+          "url": "https://github.com/srotya/sidewinder-grafana/"
+        }
+      ]
+    },
+    {
       "id": "grafana-kairosdb-datasource",
       "type": "datasource",
       "url": "https://github.com/grafana/kairosdb-datasource",


daniellee:
Thanks for submitting this. I'm on vacation but will back next week and will check it out then.

ambud:
Thank you @daniellee

daniellee:
Finally got some time to look at this. Here's my feedback so far:

- files in the dist directory need to be committed to Git. The grafana server cannot build plugins at runtime - it just loads them.
- The readme needs to be improved. A quick intro to Sidewinder (like you have in Github readme for Sidewinders) 
- A few screenshots would be a bonus
- The [plugin id](https://github.com/srotya/sidewinder-grafana/blob/master/src/plugin.json#L3) does not follow the [convention](http://docs.grafana.org/plugins/developing/code-styleguide/#plugin-json-mandatory). Should be either `sidewinder-datasource` or `sroyta-sidewinder-datasource`.
- The included src files look like they are generated files and not the original files? See this [line](https://github.com/srotya/sidewinder-grafana/blob/master/src/query_ctrl.js#L29) for example. When I tried to build it, I could not get it to work.

Looking forward to testing out your new TSDB. Can you tell me a bit more about it? What are you using it for?

ambud:
Thank you @daniellee for your feedback! I am converting your feedback to a checklist and marking the ones that are complete as I fix them.

About the database: 

**Solving problems with Databases:**
- Native TSDB to increase performance for read and writes (most TSDBs use an external K/V big table to store data)
- Was struggling with TSDBs to keep up with the write throughput in my previous company; single node performance couldn't exceed 200-300k/s
- Wanted something to be well integrated with the analytics frameworks as well (Spark etc.)

**Tech:**
Built this database for storing timeseries data at wire speed (fast). It uses a simple timeseries block compression system and memory mapped IO so it can write data really fast without compromising read performance. Think of something functioning like Kafka but is a database for timeseries. All metadata is always kept in memory for fast traversal. This database is NOT idempotent as its catered towards IoT use cases.

- [ ] files in the dist directory need to be committed to Git. The grafana server cannot build plugins at runtime - it just loads them.

- [ ] The readme needs to be improved. A quick intro to Sidewinder (like you have in Github readme for Sidewinders)

- [ ] A few screenshots would be a bonus
The plugin id does not follow the convention. Should be either sidewinder-datasource or sroyta-sidewinder-datasource.

- [ ] The included src files look like they are generated files and not the original files? See this line for example. When I tried to build it, I could not get it to work.

daniellee:
I see that you missed one point in your checklist:

> The plugin id does not follow the convention. Should be either sidewinder-datasource or sroyta-sidewinder-datasource.

ambud:
@daniellee sorry about that, I missed that.

daniellee:
@ambud did you commit that change with the plugin id? Don't see it [here](https://github.com/srotya/sidewinder-grafana/blob/master/dist/plugin.json#L3).

ambud:
@daniellee sorry I didn't fix the commit version; updated the PR with the commit version https://github.com/srotya/sidewinder-grafana/commit/545efb91bc7b031df6ed25b42a5fe55f8686f7fb

daniellee:
The plugin id needs to be updated in the plugin.json file in dist directory as well (grafana reads in the plugin id from the plugin.json file in the dist directory if a dist directory exists).

Can you update [this line](https://github.com/srotya/sidewinder-grafana/blob/master/dist/plugin.json#L3)  and [this line]() to `sidewinder-datasource'.

And update [this line](https://github.com/srotya/sidewinder-grafana/blob/master/src/plugin.json#L3) to `sidewinder-datasource' (remove the grafana part).

ambud:
@daniellee sorry about that; updating both the plugin repo and data source repo.

[Dist](https://github.com/srotya/sidewinder-grafana/blob/master/dist/plugin.json#L3), [Src](https://github.com/srotya/sidewinder-grafana/blob/master/src/plugin.json#L3)

And updated the PR with the plugin id. Please let me know if it works.

daniellee:
@ambud got it up and running now. Thanks for your patience!

Here's my feedback so far:

### Meta Data

Everything looks fine except for one thing - we don't allow html when importing the README.md file from GitHub. This breaks your image links. Can you resize your images with an image editing program to be 400px in height and use markdown links instead.

### Plugin

- In the data source settings, a placeholder in the Http Settings -> Url field would help first time users. You can do this using the suggest url property. In the config.html, add something like this:

    ```
    <datasource-http-settings
	    current="ctrl.current"
	    suggest-url="http://localhost:8080/db_name">
    </datasource-http-settings>
    ```

- In the query editor, the first time I click on the measurement/field segment it returns an empty list. If I type something, click outside the segment then it returns a list of measurements. In the docs, you refer to this: `FIELD name suggestion simply type measurementname in the "metric name or tag query field"` but you could not just show the list of fields/measurements from the start?

- There are some concepts in the query editor that need some help text/docs. The Window field should probably have something similar to the "Group by time" field in the InfluxDB query editor:
    
    image

- Also, the correlations concept and the auto-correlate need a help section. Couldn't find anything in the docs either.

- The Window field needs to be wider:
    image

All in all, it looks pretty good. Great work!

ambud:
Thanks a lot for your time and input!

- [ ] The Window field needs to be wider
- [ ] There are some concepts in the query editor that need some help text/docs. The Window field should probably have something similar to the "Group by time" field in the InfluxDB query editor

**Issue:**
Everything looks fine except for one thing - we don't allow html when importing the README.md file from GitHub. This breaks your image links. Can you resize your images with an image editing program to be 400px in height and use markdown links instead.

**Answer:**
Shrunk images and updated .md

Plugin

**Issue:**
In the data source settings, a placeholder in the Http Settings -> Url field would help first time users. You can do this using the suggest url property. In the config.html, add something like this:

<datasource-http-settings
    current="ctrl.current"
    suggest-url="http://localhost:8080/db_name">
</datasource-http-settings>

**Answer:** - FIXED
Updated code to add default URL

**Issue:**
In the query editor, the first time I click on the measurement/field segment it returns an empty list. If I type something, click outside the segment then it returns a list of measurements. In the docs, you refer to this: FIELD name suggestion simply type measurementname in the "metric name or tag query field" but you could not just show the list of fields/measurements from the start?

**Answer:**
That was a bug induced in a recent release (#55), the fix was released however, seems like the Docker image wasn't updated with the correct version, causing issues for you.
The server was running into an NPE for the measurment auto-complete, if you use the latest version of the Docker image, this shouldn't be an issue.

**Issue:**
Also, the correlations concept and the auto-correlate need a help section. Couldn't find anything in the docs either.

**Answer:** - FIXED
I will add some docs on auto-correlate feature on the project site. However, auto-correlate feature currently is not enabled on the server side.
Essentially, it will try to find all series that have a high degree of similarity with this one and return it along with the base.

All in all, it looks pretty good. Great work!

Thank you for the valuable feedback, early stages of the project.
On a side note, unlike other data sources, the plugin doesn't have much smarts built into it at the moment, rather relies on the
server to have functionality enhancements. Relying on this anti-pattern for now since I am rapidly adding functionality to the project,
it would be difficult for the users to frequently update the Grafana plugin. Once the initial MVP stabilizes, I will come back for a
second iteration of the plugin to improve the functionality.

ambud:
@daniellee I updated the plugin to add the recommendations you have suggested.

daniellee:
The metric-segment-model component is really hard to work with. I sent a PR with some improvements so that the measurement dropdown shows values without having to type something:

https://github.com/srotya/sidewinder-grafana/pull/1

ambud:
Thanks for the contrivution @daniellee, I have merged your commit and update the repo.json (PR)

ambud:
@daniellee please let me know if this can be merged?

daniellee:
Excuse the delay - a combo of catching up after vacation and a getting a cold.

Everything looks good. Published here: https://grafana.com/plugins/sidewinder-datasource 

When you do future updates, it should be a much faster process to get it published. Thanks for all your work on making the improvements and for taking the time to submit this.

ambud:
thanks a lot for merging this @daniellee hope you get well soon!

PR #150, opened 2017-08-07
Adds AppDynamics Datasource

dlopes7:
This datasource is developed by the AppDynamics community, it brings metrics from an AppDynamics controller.

### DIFF
--- a/repo.json
+++ b/repo.json
@@ -420,8 +420,8 @@
       "url": "https://github.com/dlopes7/appdynamics-grafana-datasource",
       "versions": [
         {
-          "version": "1.0.0",
-          "commit": "d85003813b06cf7f8cc9fd6dcffeb774ccb0ce40",
+          "version": "1.0.1",
+          "commit": "4495815872415a5565faab06a233eeca277b5269",
           "url": "https://github.com/dlopes7/appdynamics-grafana-datasource"
         }
       ]
--- a/repo.json
+++ b/repo.json
@@ -421,7 +421,7 @@
       "versions": [
         {
           "version": "1.0.2",
-          "commit": "de50250916cd3d3f9b1b7e371d3bac2ceb20c934",
+          "commit": "ebfcfbc5b3e648cbb251df20dfc8ba4f9afd485c",
           "url": "https://github.com/dlopes7/appdynamics-grafana-datasource"
         }
       ]
--- a/repo.json
+++ b/repo.json
@@ -415,6 +415,18 @@
       ]
     },
     {
+      "id": "grafana-appdynamics-datasource",
+      "type": "datasource",
+      "url": "https://github.com/dlopes7/appdynamics-grafana-datasource",
+      "versions": [
+        {
+          "version": "1.0.0",
+          "commit": "db0e1bab2c0c33375f60ebdd7291fd2010cc5705",
+          "url": "https://github.com/dlopes7/appdynamics-grafana-datasource"
+        }
+      ]
+    },
+    {
       "id": "grafana-simple-json-datasource",
       "type": "datasource",
       "url": "https://github.com/grafana/simple-json-datasource",
--- a/repo.json
+++ b/repo.json
@@ -420,8 +420,8 @@
       "url": "https://github.com/dlopes7/appdynamics-grafana-datasource",
       "versions": [
         {
-          "version": "1.0.3",
-          "commit": "1a9b998609af06faf98cd1cae564dc4ef78e960d",
+          "version": "1.0.4",
+          "commit": "b611c709414677a47a3bd1da372c7d986e8cf859",
           "url": "https://github.com/dlopes7/appdynamics-grafana-datasource"
         }
       ]
--- a/repo.json
+++ b/repo.json
@@ -420,8 +420,8 @@
       "url": "https://github.com/dlopes7/appdynamics-grafana-datasource",
       "versions": [
         {
-          "version": "1.0.1",
-          "commit": "4495815872415a5565faab06a233eeca277b5269",
+          "version": "1.0.2",
+          "commit": "de50250916cd3d3f9b1b7e371d3bac2ceb20c934",
           "url": "https://github.com/dlopes7/appdynamics-grafana-datasource"
         }
       ]
--- a/repo.json
+++ b/repo.json
@@ -415,13 +415,13 @@
       ]
     },
     {
-      "id": "grafana-appdynamics-datasource",
+      "id": "dlopes7-appdynamics-datasource",
       "type": "datasource",
       "url": "https://github.com/dlopes7/appdynamics-grafana-datasource",
       "versions": [
         {
           "version": "1.0.0",
-          "commit": "db0e1bab2c0c33375f60ebdd7291fd2010cc5705",
+          "commit": "69a33062bb560df8302190e92051b3270329769b",
           "url": "https://github.com/dlopes7/appdynamics-grafana-datasource"
         }
       ]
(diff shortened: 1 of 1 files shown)


daniellee:
Thanks for submitting this plugin! Before I start reviewing it, could you fix the plugin id. It does not follow the [plugin id conventions](http://docs.grafana.org/plugins/developing/code-styleguide/#plugin-json-mandatory). 

The convention is `[github username/org]-[plugin name]-[datasource|app|panel] `. The org part of the plugin id is incorrect, it should probably be `dlopes7-appdynamics-datasource`. 

Can you update the PR here with the correct id as well as in the two plugin.json files, [here](https://github.com/dlopes7/appdynamics-grafana-datasource/blob/master/dist/plugin.json#L3) and [here](https://github.com/dlopes7/appdynamics-grafana-datasource/blob/master/src/plugin.json#L3).

Thanks!

dlopes7:
Thanks for the comments, I've fixed the plugin ID.

daniellee:
Here's my feedback so far:

### Metadata

- Incorrect license link in the plugin.json `https://github.com/grafana/simple-json-datasource/blob/master/LICENSE`
- Why do you have a different README.md in the dist folder? Could you include the same README.md that is in the root instead (which has a good step-by-step guide to getting started)?

Just started setting up test data for AppDynamics so will get back to you with more feedback shortly.

daniellee:
Took a while to figure out how to authenticate with basic auth (basically I gave up trying to use the default user and just created a new test account). Anyway, maybe this is easier if you use on-prem AppDynamics?

Feedback so far on the plugin:

- In the query editor, the input fields are quite hard to use. I have to start typing before any applications appear in the dropdown. Same thing for the metrics field. In the metrics field, once I have selected the first part of the path (e.g. `Errors|`) I have to press Ctrl+space to get suggestion for the next part. Maybe you could look at similar query editors (graphite for example) to get an idea of how you could solve it in a better way. Maybe you could take a look at the new `gf-form-dropdown` component ([example here](https://github.com/grafana/simple-json-datasource/blob/master/src/partials/query.editor.html#L13-L17)). It only got added to Grafana recently so if you want to support older versions of Grafana then you should use the `metric-segment` component (it is harder to use than the `gf-form-dropdown` component) that is used in most query editors -> here is the graphite query editor:

    image

- Error handling needs to be improved. As an example, if I click on the eye symbol to hide a query then that usually throws an exception and the data does not get hidden.

- I could not figure out what the Segments option is used for. Maybe a help section is needed?
    image

Let me know if anything is unclear or if you need help finding examples.

dlopes7:
Awesome comments, I'm impressed that you got an AppDynamics server up that fast.

Thanks for the graphite example, I was really looking on how to solve that issue.

About the SaaS authentication, it should be the same, unless the user you were using did not have access to the REST api, I'll document that better as well

I'll work on these and update the PR request.

dlopes7:
The README.md being different is a mistake on my build script.

dlopes7:
@daniellee I was unable to come up with a better way to work with Metric autocomplete (having to press Ctrl+Space). Ideally I would have wanted the field to update whenever text changes, even if it wasn't the user that typed in the change (when he clicks on one of the options), but I was unable to do so, or to find an example of someone doing that.

I've updated the **Application** field to use `metric-segment`, but it doesn't work for the **Metric** field because the new text overwrites whatever was in the input before, so I am leaving as "ctrl+space" or a new letter being typed for now, I at least changed the min-length of the Metric field to zero, so the user doesn't have to type anything to get suggestions.

Link to license was wrong and was fixed. README.md wasn't included on my gulp script, now it is.

Added error handling for some different cases, and the functionality to hide the metric when the user clicks the eye icon.

Added a helper text for the Segments option.

Would you kindly review the Datasource? If the metric auto-complete is unacceptable I would love to get some ideas on how to fix this.

Thanks!

daniellee:
Thanks for your patience. I'm back from vacation now.

I like the changes and I think it is good enough to publish.

If you are up for a challenge, I think the autocomplete could be improved. It is a lot of work but you could do the auto complete like in the [graphite query editor](https://github.com/grafana/grafana/blob/master/public/app/plugins/datasource/graphite/query_ctrl.ts). 

image

Did a quick prototype and it looked like this:

image

The html would look like this:

```html
<label class="gf-form-label width-8">Metric</label>
<div ng-repeat="segment in ctrl.segments" role="menuitem" class="gf-form">
  <metric-segment segment="segment" get-options="ctrl.getMetricNames($index)" on-change="ctrl.segmentValueChanged(segment, $index)"></metric-segment>
</div>
```

The getMetricNames function would be something like this:

```js
this.getMetricNames = (index) => {
    return this.appD.getMetricNames(this.target.application, this.getSegmentPathUpTo(index))
    .then(this.transformToSegments(false));
};
```

```js
getSegmentPathUpTo(index) {
  const arr = this.segments.slice(0, index);
  let segments = '';
  for (let i = 0; i < arr.length; i++) {
    segments += arr[i].value;
  }
  return segments;
}
```

```js
this.segmentValueChanged = (segment, segmentIndex) => {
  //if last segment in query then add a new field
  this.segments.push(this.uiSegmentSrv.newSelectMetric());
  //else check other segments
};
```

What do you think? I can publish the plugin as it is.

dlopes7:
@daniellee You are good with this stuff! I really like your ideas, lets hold for that publish for a while then.

I'll try to implement it this way, if you have an example of how to delete one of those segments, that would be awesome too.

Thanks for the effort with your example

dlopes7:
Thanks for the tips Daniel, I've modified the Metric autocomplete feature, I think it works much better now.

daniellee:
That looks much better. One more step and then I think this is ready to publish. In the Graphite data source, when I change a previous query part then it removes the query parts after it. An example:

1. Select a metric:

image

2. But I want to change it and instead of "Service Endpoints", I want to select a metric from "Backends":

image

3. When I select "Backends" then the 3 query parts ("Rest Services", "/SampleApp/sql_error", "Errors Per Minute") should disappear and it should look like this:

image

Which I hope answers your question on how delete segments. (Sorry about the delay, got hit by manflu but I'm better now.)

dlopes7:
I did not want to implement that because I would lose one functionality, the user can change a segment to '*' to get all metrics that match, in that case I did not want to delete the segments after it.

I ended up implementing it with a conditional, will delete the others if the value of the one being edited is different than '*', thanks!

daniellee:
Excuse, the delay. Snowed under at the moment. 

I tested this again and now it is very close. I found one thing that has to be fixed first and then I can publish this. 

The showstopper is that when I reload a dashboard it does not seem to load the metric segments when opening the query editor:

image

There are also a few improvements that could be made too :-)

- Could you set a default value for the Segments field (1 perhaps?) otherwise I sometimes got an error message before I had a chance to fill anything in there.
- In the query editor, use the css class `query-keyword` on labels. Otherwise the label and value look too similar:
    
    image

- The Segment/Fullpath dropdown is a different height from everything else on the row. Wrap it in a div with a `gf-form` css class.

- There is a gap between the last segment and the legend fields:
    image
    You can fix this using the grow css classes. Have a look at the Elasticsearch data source for an example. It should be something like this: 
    ```
    <div class="gf-form gf-form--grow">
	  <label class="gf-form-label gf-form-label--grow">
	  </label>
    </div>
    ```

dlopes7:
Thanks for the detailed review!

I've implemented the fixes, and it looks much better now.
The metric segments not copying and disappearing on reload has been fixed as well.

fixes

daniellee:
Merged and published! Great work!

PR #151, opened 2017-08-08
Add Skydive datasource

safchain:


daniellee:
Thanks for submitting this plugin! Looks like a panel that people will want to add to their Grafana dashboards.

Before I start reviewing it, could you fix the plugin id. It does not follow the [plugin id conventions](http://docs.grafana.org/plugins/developing/code-styleguide/#plugin-json-mandatory). 

The convention is `[github username/org]-[plugin name]-[datasource|app|panel] `. In your case, you can skip the org part so it should probably be `skydive-datasource`. 

Can you update the PR here with the correct id as well as in the two plugin.json files, [here](https://github.com/skydive-project/skydive-grafana-datasource/blob/master/dist/plugin.json#L3) and [here](https://github.com/skydive-project/skydive-grafana-datasource/blob/master/plugin.json#L3).

I noticed that you have the plugin.json in the root. If you want to simplify file paths, then you could move the plugin.json to the src directory. Grafana will only load the plugin.json in the dist directory so it does not matter where the source plugin.json is located. You could then change `dist/src/img` to be just `dist/img`.

safchain:
Thanks for the review. I made the changes, changed the plugin id and move the plugin.json and the img folder as well

daniellee:
This image is 0 bytes which causes an error when trying to publish: https://github.com/skydive-project/skydive-grafana-datasource/blob/master/dist/img/skydive-logo.png

Also noticed that you still have `dist/src/img` directory which I think you don't need anymore?

safchain:
Sorry for that, forgot to submit the last changes :( which now done

daniellee:
Thanks for your patience (was on vacation last week). Took me a while to get Skydive up and running but it was easy after I found the official docker-compose file.

Here's my feedback:

### Metadata

- Readme looks fine. Nice screenshot and had enough links (gremlin, docs) to get started with.
- plugin.json is correct

### Plugin

- No bugs that I could find :+1: 
- Simple query editor which works as it should. A small visual thing is that the labels for Query and Title don't line up (see below):

image

Great work! Published and merged: https://grafana.com/plugins/skydive-datasource

P.S Just curious but is there any way to alias multiple flows (if it is a single flow then the Title field works). For example, the following query: `G.Flows().Sort()` returns a lot of flows with guids. In Graphite, you can alias by another field to get more human readable entries in the legend.

image

safchain:
Hi @daniellee, thanks a lot for testing and merging ! Good you've been able to test Skydive too :)

For the aliases when multiple flows, I will add this in the next release.

PR #152, opened 2017-08-09
add stackdriver datasource

mtanda:


daniellee:
Hi mtanda!

I haven't started reviewing this yet but I noticed that the plugin.json file refers to a screenshot that is not in the repo: https://github.com/mtanda/grafana-google-stackdriver-datasource/blob/master/dist/plugin.json#L23

Looking forward to testing this out!

mtanda:
Thanks for quick check!

I add a screenshot of query editor and update README.

If you are not familiar with Stackdriver, please read [API doc](https://cloud.google.com/monitoring/api/ref_v3/rest/v3/projects.timeSeries/list).
The Stackdriver is little similar to Prometheus, it has `filter`, `reducer` and `aligner`.
- `filter` is [label matcher](https://prometheus.io/docs/querying/basics/#time-series-selectors).
- `reducer` is [aggregation](https://prometheus.io/docs/querying/operators/#aggregation-operators).
- `aligner` is [aggregation over time](https://prometheus.io/docs/querying/functions/#<aggregation>_over_time()) in Prometheus.

I'm waiting for your feedback about usability.
Please enjoy Stackdriver!!

daniellee:
Please excuse the delay - vacation times. I've started reviewing the plugin and here is my feedback so far:

## Metadata

- The plugin.json looks fine
- The README is pretty good - it would have been hard to figure out everything without it.

## Plugin

- The [testDatasource](https://github.com/mtanda/grafana-google-stackdriver-datasource/blob/master/src/datasource.js#L135-L145) function needs a catch function. If it does not work then the only error you see is "Unknown error". Otherwise the config section looks good.
- Query editor: it does work but maybe the filter section could be improved? Having to write a filter like: `metric.type = "compute.googleapis.com/instance/cpu/usage_time"` is not very discoverable. Is it possible to make that into a dropdown with a list of metric types? The Stackdriver Metric Explorer has a dropdown with metrics (see below). Is that possible via the API?
image
- Query editor: how does the alias field work? Am I doing something wrong here?
image
- Query editor: The align field was a little bit hard to use. It seems to only accept seconds (like "60s). How does the auto functionality work here?
image

Overall, this is a great start. If you could make the filter part more user friendly then this is ready to publish. Thanks for your patience!

daniellee:
Another question. How does the authentication work? If I setup a Stackdriver data source, will other users be able to use it or do they have to authenticate as well?

mtanda:
Thanks for feedback!

> The testDatasource function needs a catch function. If it does not work then the only error you see is "Unknown error". Otherwise the config section looks good.

I should set `required` for client id. Improve testDatasource() little...

> Query editor: it does work but maybe the filter section could be improved? Having to write a filter like: metric.type = "compute.googleapis.com/instance/cpu/usage_time" is not very discoverable. Is it possible to make that into a dropdown with a list of metric types? The Stackdriver Metric Explorer has a dropdown with metrics (see below). Is that possible via the API?

It is possible.
Honestly, I want to improve query editor.
I think Ace editor integration is good option, try to use it.

> Query editor: how does the alias field work? Am I doing something wrong here?

reducer should be `REDUCE_NONE`, if you want to display each instance metrics and show instance name as legend.
And specify `labels` not `label`.
So, correct alias pattern is `{{metric.labels.instance_name}}`

> Query editor: The align field was a little bit hard to use. It seems to only accept seconds (like "60s). How does the auto functionality work here?

Sorry, I forgot to implement `auto`...
Set `intervalMs` as default.

mtanda:
> Another question. How does the authentication work? If I setup a Stackdriver data source, will other users be able to use it or do they have to authenticate as well?

The answer is `do they have to authenticate as well`.

Google provide another authentication method, `Service Account`.
https://cloud.google.com/monitoring/api/authentication

If this datasource plugin support it, no need to extra authentication.

But we can't use it from JS in Browser. (no document about using Service Account from JS...)
https://cloud.google.com/monitoring/docs/reference/libraries
https://developers.google.com/api-client-library/javascript/start/start-js

Maybe nodejs library support it...
https://github.com/GoogleCloudPlatform/google-cloud-node

mtanda:
I improve the query editor, please check!
https://github.com/mtanda/grafana-google-stackdriver-datasource/pull/2

daniellee:
@mtanda I had a quick look at your changes and I'd like to show this to the rest of Grafana Labs when they come visit us here in Stockholm next week. I think that we should help you with this plugin.

I apologize for me being so slow at providing feedback, but hopefully we can get this published next week.

mtanda:
I don't mind :)
I know you are vacation time.

I notice you create Azure plugin recently, so if this Stackdriver plugin is published, Grafana is ready for all major Cloud service.

I'm looking forward to Grafana Labs member feedback.

kerneljack:
Hi I'm having trouble getting the datasource to connect to Stackdriver, are you able to help @mtanda @daniellee? 

I've created an issue for it here if you'd like more details, thanks:
https://github.com/mtanda/grafana-google-stackdriver-datasource/issues/5

arbarlow:
Any word on this? I'm happy to help as my company has a vested interest in this datasource at so I would pitch in!

mtanda:
Sorry for late comment here.
I investigate the issue in https://github.com/mtanda/grafana-google-stackdriver-datasource/issues/5 .
If anyone has same problem, please watch the issue in my datasource plugin repos.

mattayes:
@daniellee Any update from the Grafana Labs team?

lepatryk:
@daniellee we are using it internally at Firebase. It's a well written plugin with great user experience (kudos to Mitsuhiro). We would love to see it officially included in Grafana Labs so that we can recommend it to our customers.

What else is missing? I'm happy to make the changes if necessary.

timc3:
Is there a way of making the datasource authentication persistant in some way? Currently having to login periodically.

vijayjt:
@mtanda With respect to the question around authentication. I came across this [post](https://stackoverflow.com/questions/28751995/how-to-obtain-google-service-account-access-token-javascript) on Stackoverflow; could you use that in order to add support for using service accounts?

mtanda:
I didn't test service accounts.
But it may work in recent version.
https://github.com/mtanda/grafana-google-stackdriver-datasource/pull/24

vijayjt:
@mtanda Great thanks. I hadn't noticed that feature has been added. I'll give that a go.

mtanda:
Close this plugin add request and stop development of my plugin because Grafana Labs start to develop their plugin.
And, they don't want to have two stackdriver plugins.

https://github.com/grafana/grafana/tree/stackdriver-plugin/public/app/plugins/datasource/stackdriver

PR #154, opened 2017-08-18
add new sni-pnp-datasource releases

sni:
this PR adds the latest two releases of the pnp datasource.


daniellee:
Published! Nice to see that you added templating support.

PR #155, opened 2017-08-22
update gauge panel to 0.0.4

briangann:
Added new feature for mapping tick values to text (180 degrees = North, etc)

Also includes a couple of bug fixes causing crashing when updating limits.

Thanks!

Brian


daniellee:
Both plugin updates published. Nice work! The tick mapping is very slick:

image

briangann:
many thanks, more features are being implemented too :)

PR #157, opened 2017-09-01
Adds GeoLoop Panel

samhatchett:


daniellee:
This looks really cool. I'm a bit snowed under this week but will get back to you with feedback next week.

samhatchett:
cool, thanks! I should have a more immersive example by late next week as well, and so will update the readme once that's ready.

samhatchett:
i've updated the plugin slightly with a fuller demo (including animated gif) and slightly expanded set of geo parsing options. please let me know if you have any questions!

daniellee:
Sorry, snowed under with work at the minute. Some quick feedback:

- image links for the [geoloop-icon.png](https://github.com/CitiLogics/citilogics-geoloop-panel/blob/master/dist/plugin.json#L14-L15) in the plugin.json is incorrect. There is an extra `src` that shouldn't be there.
- The README looks fantastic!
- I will ask our designer to make a logo for you
- The options tab looks neat and tidy.

Any tips on finding some sample geoJSON data?

samhatchett:
Thanks so much! Glad to get the feedback.

For the icon image URLs, these are valid relative to the root repository directory. Are these supposed to be relative to `dist`? And also same question the "screenshots" image URLs.

As for getting GeoJSON data, I usually end up converting from shapefiles - but there are simple online generators if the features are not too complicated. I will add a few words in the Readme about this topic.

daniellee:
The urls are relative to the plugin.json file. So if the plugin.json is in the root of src and dist, then the url should be something like this: `images/geoloop-icon.png`.

daniellee:
Might be a good idea to include a sample geojson file and maybe some instructions on how to use something like [Caddy](https://caddyserver.com/tutorial) to serve one file (it is super easy with Caddy so I should add instructions for this for WorldMap too).

daniellee:
- Getting an error on this line if the field name does not match: https://github.com/CitiLogics/citilogics-geoloop-panel/blob/master/src/geoloop_ctrl.js#L311 Need to add a check if it is null or undefined.
- Took me a while to figure out that the alias was expected to contain both the label and the value. This means this will not work with Graphite and also that I should not add anything to the ALIAS BY field. You do have something in the docs for this but could you make it more explicit and say that the alias must be in this form. You could get this to work for Graphite if you wanted to by skipping the regex matching and just using the value.
- Also took me a while to figure out the path to the GeoJSON ID Path field needs to be something like `properties.my_id`. More docs!
- "Allow pan/zoom" feature didn't seem to do anything for me.

So basically more docs and examples with GeoJSON please. Otherwise this looks great. I could publish this now but I think without a better Getting Started guide with an example GeoJSON then people will get stuck on the setup.

samhatchett:
regarding the `images` path, i see now in the Grafana docs: (emphasis mine)

> When Grafana starts it will scan the plugin folders and mount every folder that contains a plugin.json file __unless the folder contains a subfolder named dist__. In that case grafana will mount the dist folder instead.

... so that explains my confusion. There are 2 `plugin.json` files, but only one matters, and the URLs are relative to the one in `dist`. Got it.

And thanks __so much__ for the review. I'm not a Graphite user, and so I'm not totally tracking what would be required to make that work. However I will put it on my list to figure out, perhaps for the next rev.

 I will circle back on this - update docs and add a sample.geojson file, fix the error you mentioned. Might take about a week.

/

daniellee:
Our designer, Trent came up with this icon:

image

samhatchett:
Great icon - thanks Trent! 

I've addressed all the issues we've discussed. Please look over the new Readme and let me know if you still think anything is missing or not yet clear enough.

daniellee:
I apologize for the long delay. Your changes look great! Merged and published: https://grafana.com/plugins/citilogics-geoloop-panel

PR #163, opened 2017-09-05
Update all btplc plugins

benthorner:
The following changes have been made:

   - Improve layout of options tabs (like Grafana default panels)
   - Add link selector for trend-box and alarm-box panels
   - Move panel link into HTML for trend-box and alarm-box panels
   - Add individual dot links feature to the status-dot panel

### DIFF
--- a/repo.json
+++ b/repo.json
@@ -3,110 +3,110 @@
     {
       "id": "btplc-trend-box-panel",
       "type": "panel",
-      "url": "https://github.com/BTplc/grafana-trend-box",
+      "url": "https://github.com/BT-OpenSource/bt-grafana-trend-box",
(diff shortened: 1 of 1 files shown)


benthorner:
We are also moving all BT's Open Source project a new account, so the URLs have changed to reflect this. All open issues have been resolved, so there is no reason to keep the legacy repos.

daniellee:
The drilldown link support is very nice. Great work as usual. All plugins merged and published.

benthorner:
Thanks Dan :-) When I upgraded our end I spotted an issue with the plugin.json links and a small bug with the drilldown link default value. I've made a new pull request to get these fixed :-|

PR #166, opened 2017-09-14
gnocchi: release version 1.5.5

sileht:


daniellee:
Merged and published.

Just curious (and also want to write about the new version in the our weekly blog update), but you could you explain how the new reaggregation feature works. Thanks!

sileht:
Gnocchi does pre-computation of timeseries. So you describe what kind of retention and aggregation method you want (ex: aggregate the mean every 10 minute for 1 year). Some processes compute this is background when new raw datapoints are available. Then when the user (Grafana here) queries Gnocchi, the timeseries are already computed. It's very fast.

Then Gnocchi API also allows to (re)aggregate timeseries. And because stored timeseries are already aggregated, we call this reaggregation.

The use case behind is: you aggregate the mean of cpu utilization on 1 minutes. Then later you want the mean of all your servers at once. So you "reaggregate" them. You got the mean of mean of all cpu utilization. Of course not all combination of aggregation/reaggregation make sense.
Another example: with collectd you record the mean of  sleep processes, running processes, zombie, uninterrupted, ... . You can make a graph that show the sum of all mean of these processes metrics to have the total number of processes.

daniellee:
Thanks for the explanation. Reaggregations that don't make sense is a general problem for most data sources, (summing averages/means in Graphite etc.).

PR #168, opened 2017-09-19
Update stagemonitor

trampi:


daniellee:
Why so many versions? I can only see two commits that seem to have any actual changes.

trampi:
To be honest, I have expected this question. There are two reasons for that:

1) We have automated the release process, because of reasons number two.

2) We want to make things simple for users of stagemonitor. If someone is using stagemonitor 0.31.0, he should not need to research, if 0.30.0 of the stagemonitor grafana plugin contains dashboards compatible with 0.31.0. We also want to avoid to have "compatibility tables", as they do enhance complexity for our users. Even if the dashboards have not changed, the user might be disappointed or confused when installing the grafana plugin and see no plugin for e.g. stagemonitor version 0.31.0.

daniellee:
OK, no worries. Just making sure that it was intentional :-)

Published all the versions except for "0.80.0.RC1" - our api for Grafana.com wouldn't accept the version number in that form. Is it ok to skip that version?

trampi:
Hi daniellee,
thanks for the feedback :-)

I updated the PR and removed the RC version - this should not be used anyway.

Thanks for your time!

PR #169, opened 2017-09-19
OpenNMS Helm 1.0.0, successor to the OpenNMS Data Source

j-white:
We've migrated our previous data source to a full fledged application that now provides 2 data sources and 2 custom panels.

### DIFF
--- a/repo.json
+++ b/repo.json
@@ -596,6 +596,11 @@
       "url": "https://github.com/OpenNMS/grafana-opennms-datasource",
       "versions": [
         {
+          "version": "2.1.1",
+          "commit": "b5ff284de5454a5ab1bf6b154538e956cfbe0d96",
+          "url": "https://github.com/OpenNMS/grafana-opennms-datasource"
+        },
+        {
           "version": "2.1.0",
           "commit": "23866f0d3fee26d084bd54e7333f21ae00b98d63",
           "url": "https://github.com/OpenNMS/grafana-opennms-datasource"
@@ -614,6 +619,18 @@
           "version": "2.0.0",
           "commit": "c089ce7c2117f709ef4c944a332776dd290055af",
           "url": "https://github.com/OpenNMS/grafana-opennms-datasource"
+        }
+      ]
+    },
+    {
+      "id": "opennms-helm",
+      "type": "app",
+      "url": "https://github.com/OpenNMS/opennms-helm",
+      "versions": [
+        {
+          "version": "1.0.0",
+          "commit": "c6a950cc5d965ab4071b7eab556d3d60e791e971",
+          "url": "https://github.com/OpenNMS/opennms-helm"
         }
       ]
     },


j-white:
Any feedback here?

We'll also be providing .rpm/.deb packages for this plugin and was hoping to get this reviewed before we publish these.

daniellee:
Haven't had time this week to review any plugins - working hard on Grafana 5.0. I will have time next week. Sorry about the delay.

pbrane:
Still looking for that review. Hopefully this week?

daniellee:
Some quick feedback:

- The dist subdirectory needs to be committed. Grafana does not have any way of building plugins on the fly.
- The metadata looks fine (plugin.json, README.md)
- For the query editors, they look nicer if you use a grow label on each row after the last field:
    ```html
    <div class="gf-form gf-form--grow">
        <div class="gf-form-label gf-form-label--grow"></div>
    </div>
    ```
    Then it looks like this (the grow label fills in the space between the + and the hamburger icon):
    image
- The css for the options for the two panels looks good.

I'll get back to you with more feedback once I have a minion up and running to test with.

pbrane:
FYI: You don't *need* a Minion to test this, btw.

j-white:
Thanks for the input. You should be able to point the data sources to our Demo system at `https://demo.opennms.org/opennms` with username `demo` and password `demo` if you want to test it out. That should save you from having to bootstrap and OpenNMS stack.

j-white:
We do commit the `dist/` folder in the `master` branch. I'll make sure any future commits we reference include this.

j-white:
I've updated the hash to reference a version that contains the `dist/` directory.

I also looked into adding the `grow` tags, but they don't appear to be support on the [gf-form-input](https://github.com/OpenNMS/opennms-helm/blob/v1.0.0/src/datasources/perf-ds/partials/query.editor.html#L10) classes of the [metric-segment](https://github.com/OpenNMS/opennms-helm/blob/v1.0.0/src/datasources/fault-ds/partials/query.html#L10) directives.

daniellee:
@j-white you don't use them directly with the metric-segment directives. An example. Here is the first row in the faults query editor:

```html
<div class="gf-form-inline">
        <div class="gf-form">
            <label class="gf-form-label query-keyword width-7">SELECT</label>
            <label class="gf-form-label width-7">all alarms</label>
        </div>
    </div>
```

You would add the grow label after the last `gf-form`div:

```html
    <div class="gf-form-inline">
        <div class="gf-form">
            <label class="gf-form-label query-keyword width-7">SELECT</label>
            <label class="gf-form-label width-7">all alarms</label>
        </div>
        <div class="gf-form gf-form--grow">
            <div class="gf-form-label gf-form-label--grow"></div>
        </div>
    </div>
```
Before:
image

After:

image

I have more feedback coming soon but it is looking good so far. You have really put a lot of work into this.

j-white:
That row I was able to figure out, I was having trouble with the others though.

Cool, looking forward to it.

daniellee:
## Plugin feedback

- Getting errors when in the Alarm Histogram if I choose the wrong data source type. Could you make the error more explicit: "The Alarm Histogram requires an OpenNMS Faults data source". This is the error you get instead:

    image

- Also, get very strange errors in the Graph panel if I try to use the OpenNMS Faults data source there. An explicit error would be better if possible.

- When using the OpenNMS Performance data source, if I duplicate a query then I get a request error "Query source label 'test' conflict: source with that label is already defined." but I have to go look for it in the Chrome console. Can this be shown using appEvents.emit:
    `appEvents.emit('alert-error', ['Error', error.message]);`

- For the modal queries and testing against demo.opennms.org, it can be very slow. But I guess that is nothing to do with the plugin code:

    image

## UX Feedback

- The modals for the OpenNMS Performance data source show rows that can be selected but that is not instantly obvious as they are missing the `pointer` css class which will turn the mouse pointer into a hand on hover. A [hover highlight](https://developer.mozilla.org/en-US/docs/Web/CSS/:hover) would be a bonus.

    image

- Also, a little bit of padding around the borders, the rows and the search input would make this modal look nicer. There are some very similar lists in Grafana so you could copy the css from there.

- The Cancel button has the wrong css class (see here for the style guide: http://localhost:3000/styleguide/buttons)

- From a UX perspective, it is hard to see that it is possible to click on the Node, Resource and Attribute labels and get a modal. Is there some sort of [affordance](https://www.smashingmagazine.com/2014/06/affordance-most-underrated-word-in-web-design/) you could add that would indicate to a user "click here"? Another way to go would be to use something like gf-form-dropdown instead of a modal but that is probably too big a change.

- As mentioned above, grow labels should be added as the last element inside a `gf-form-inline` div. Should look like this:

    image

## Conclusion

All the above points are just improvements that could be made. There is nothing that would stop me publishing the plugin. It looks like you have put a lot of work into this. Apologies that it took so long to get around to reviewing.

What do you want to do with the existing OpenNMS plugin - deprecate or leave as it is? If deprecating then it would be great if you could update the documentation with a link to the new plugin.

Is it ok to use the latest commit (c062fb1651301c11b70b01b91e5429b33740c49c) from the master branch when publishing?

j-white:
Thanks for the extensive review, we'll look at addressing the points in future iterations.

I'd opt for using the reference commit instead of the latest, since it matches the 1.0.0 tag and the same version used by our RPM packages.

The previous data source should be deprecated. The README has already been edited. Should we also bump the plugin version to reference the update README?

daniellee:
@j-white yes, can you bump the version on the old opennms repo. Thanks.

j-white:
Version has been bumped for the deprecated plugin (which includes the deprecation notice in the README)

daniellee:
Sorry for the delay - old plugin is deprecated and the new one published:

https://grafana.com/plugins/opennms-helm-app

j-white:
Awesome, thanks for the help.

pbrane:
Thanks, @daniellee

PR #171, opened 2017-09-21
Patch update to BT Status Dot panel

benthorner:
This patch removes an incorrect use of ng-if.

### DIFF
--- a/repo.json
+++ b/repo.json
@@ -71,7 +71,7 @@
         },
         {
           "version": "0.1.2",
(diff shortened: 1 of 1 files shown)


daniellee:
You are missing a double quote after the version number: https://github.com/BT-OpenSource/bt-grafana-status-dot/blob/master/src/plugin.json#L7

benthorner:
Oops. Should be fixed now. I've also updated the Alarm Box panel to use a new default to work with the table format for ES.

daniellee:
Merged and published both.

PR #172, opened 2017-09-22
Upgrade BT Box and Dot panels

benthorner:
This change includes a few patches to remove the use of CSS files.


daniellee:
All 3 updates published.

PR #173, opened 2017-09-27
update ajax-panel to 0.0.3

ryantxu:
adds an iframe option

see also #140


daniellee:
Merged and published.

PR #177, opened 2017-10-06
Update BT box/dot plugins (fix threshold bug)

benthorner:
This update fixes a bug in the box/dot panels where thresholds where being interpreted as strings instead of numbers. Apologies for the volume of pull requests (this bug wasn't caught in testing).


daniellee:
Merged and published.

PR #181, opened 2017-10-18
Publish Influx Admin 0.0.3

ryantxu:
This has a few useful updates

### DIFF
--- a/repo.json
+++ b/repo.json
@@ -204,7 +204,7 @@
       "id": "grafana-worldmap-panel",
       "type": "panel",
       "url": "https://github.com/grafana/worldmap-panel",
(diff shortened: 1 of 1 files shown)


daniellee:
Merged and published.

Was it hard to convert it to TypeScript?

ryantxu:
> Was it hard to convert it to TypeScript?

Not at all -- the only the that was hard was the [missing constructor](https://github.com/NatelEnergy/grafana-influx-admin/blob/master/hack/common.d.ts#L135)

Thanks for the template project, that is where i got stumped before.  Once converted to typescript it is much easier to catch stupid errors :)

PR #182, opened 2017-10-19
add grafana-progress-list-plugin

jonyrock:


daniellee:
Just noticed that the [plugin id in the plugin.json](https://github.com/CorpGlory/grafana-progress-list/blob/master/dist/plugin.json#L4) file does not follow the [conventions](http://docs.grafana.org/plugins/developing/code-styleguide/#plugin-json-mandatory). It should be `corpglory-progresslist-panel`

Will start reviewing this soon!

jonyrock:
@daniellee is only thing I need to do ? :)

daniellee:
@jonyrock no, I think you will need to do more changes. I just had a quick test and getting some errors when trying out different types of queries. I tested with InfluxDB like in the examples on the tutorial page but did not manage to get anything to work yet.

rozetko:
@daniellee hi! I was testing this plugin with Grafana versions 4.5.2 and 4.6.2. Plugin didn't work at 4.6.2 due to bug, which I fixed just now. Looking forward to your further feedback!

daniellee:
@rozetko @jonyrock the plugin id is still incorrect: https://github.com/CorpGlory/grafana-progress-list/blob/master/dist/plugin.json#L4 

It should be `corpglory-progresslist-panel` in the two plugin json files (src and dist).

rozetko:
Hi, @daniellee! I changed it, everything should be ok now.

PR #194, opened 2017-11-23
Bump gnocchi datasource 1.6.0

sileht:


daniellee:
Merged and published. If I understand it correctly this version adds support for dynamic aggregations?

http://gnocchi.xyz/rest.html#dynamic-aggregates

sileht:
Exactly

PR #195, opened 2017-11-27
marcuscalidus-svg-panel

MarcusCalidus:
as discussed in this forum post:
https://community.grafana.com/t/panel-with-svg-and-overlayed-values-for-iiot/2953/11


MarcusCalidus:
new PR with better version

PR #197, opened 2017-11-30
add natel-usgs-datasource

ryantxu:
Here is a datasource that lets you read data from the USGS water service servers.

I tried a more general datasource for various public services, but this works well so i think better to just publish :)

### DIFF
--- a/repo.json
+++ b/repo.json
@@ -1046,6 +1046,18 @@
       ]
     },
     {
+      "id": "natel-usgs-datasource",
+      "type": "datasource",
+      "url": "https://github.com/NatelEnergy/grafana-usgs-waterservices",
+      "versions": [
+        {
+          "version": "0.0.1",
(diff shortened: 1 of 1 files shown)


daniellee:
I apologize for the long delay in reviewing this. It looks great and it is so easy to get started. I found one small bug - if you search for site that does not exist then the screen goes dark and there is the following error in the console:

`GET https://waterservices.usgs.gov/nwis/iv/service/?format=rdb&sites=01304057abc 400 (Invalid Site number format: character is not a digit at argument 0, position 8, server=[caas01])`

It would be great if you could add a catch to that promise.

But that is only a small thing so it is published with this commit: de4060ea6bc3c1ad6e8512cd4ff6356f5d15573e

ryantxu:
Ha - I was looking forward to poking you about it tonight!

Thanks for the bug report -- I'll post an update soon

PR #199, opened 2017-12-01
Update clickhouse datasource

hagen1778:
Changes https://github.com/Vertamedia/clickhouse-grafana/blob/master/CHANGELOG.md

### DIFF
--- a/repo.json
+++ b/repo.json
@@ -1585,8 +1585,8 @@
           "url": "https://github.com/Vertamedia/clickhouse-grafana"
         },
         {
-          "version": "1.2.1",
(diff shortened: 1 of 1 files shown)


daniellee:
Great to see all the changes. Love the Ace Editor integration and the reformat query function. For the next version, there are a few visual glitches to fix:

- The style with font-size set to 1 rem means the Column:Date label's height is larger than the other elements on the same row: 
    image

- When both Show Help and generated sql are chosen then it looks a bit strange:
    image

Merged and published!

hagen1778:
Thank you for feedback! I'll do my best to fix those glitches.

PR #207, opened 2017-12-16
Fix for AppDynamics templating

dlopes7:
There was a small bug when only metrics were used on templating and not apps. This update fixes it.


daniellee:
Merged and published.

PR #211, opened 2017-12-20
Update BT status plugin

jonasbleyl:
* Fix tooltip positioning so that it is visible when embedding.


daniellee:
Merged and published.

PR #215, opened 2017-12-27
add sumologic datasource plugin

mtanda:


daniellee:
I apologize for the long delay in reviewing this. Tested on v5 and it works quite well - just one small thing to fix.

- Hiding and showing queries using the eye icon in the query editor does not seem to work.

Other improvements (just suggestions):

- The docs are bit out of date (my fault for not publishing earlier). 
- Is it possible to show the errors in the query editor (like the mysql/postgres data sources do) if the query is invalid?
- Could you add a few simple queries to the docs to make it easier to get started?
- This field in the query editor should be wider:
    image

mtanda:
@daniellee Thanks for feedback!
I update plugin README and code!!

daniellee:
Thanks! Merged and published: https://grafana.com/plugins/mtanda-sumologic-datasource

mtanda:
Thank you!

PR #216, opened 2017-12-28
Update GPA openHistorian datasource

ritchiecarroll:
This update enhances the Grafana data source plugin for the openHistorian to version 1.0.2 with new query builder screens and adds better documentation.

Note that this version also creates a `dist` sub-folder for the target distribution files where the prior versions did not, in case this is relevant to the import process.


daniellee:
Merged and published. The new dist folder should not cause any problems.

ritchiecarroll:
Thanks Daniel...

PR #218, opened 2018-01-16
Update BT peak report plugin

jonasbleyl:
* Fix readme file location in dist


daniellee:
Thanks Jonas! I am setting up automatic linting for the plugin meta data (which is how I found this) so hopefully errors like this won't get through in the future.

PR #227, opened 2018-02-05
add humio2grafana datasource

xorsnn:
This PR adds humio2grafana datasource plugin. Humio is log management without limits. Query, aggregate, and visualize your application data instantly, on-premise or in the cloud.
https://www.humio.com/

### DIFF
--- a/repo.json
+++ b/repo.json
@@ -1977,6 +1977,11 @@
           "version": "1.0.2",
           "commit": "9a18261eee1a7aed5e7bc0897ba8f2481615330d",
           "url": "https://github.com/humio/humio2grafana"
+        },
+        {
+          "version": "1.0.3",
+          "commit": "f99c907ba4e7e4f30ea8e85da512e9e1558e2d3f",
+          "url": "https://github.com/humio/humio2grafana"
         }
       ]
     }
--- a/repo.json
+++ b/repo.json
@@ -1967,6 +1967,18 @@
           "url": "https://github.com/devicehive/devicehive-grafana-datasource"
         }
       ]
+    },
+    {
+      "id": "humio-humio2grafana-datasource",
+      "type": "datasource",
+      "url": "https://github.com/humio/humio2grafana",
+      "versions": [
+        {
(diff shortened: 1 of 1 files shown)


daniellee:
Sorry for the delay in reviewing - have been busy with Grafana 5.0.

Here's my feedback so far:

- Very easy setup - really nice that the url for the humio api is prefilled.
- Is the access key supposed to be secret? If so, you should use `secureJsonData` (instead of jsonData) on the config page to save it so that it will be encrypted.
- It is not possible to delete a query in the query editor after adding it.
- The "Open query in Humio" link does not work for me.

Nice work - it was very easy to get this up and running. If you can fix the access key to be encrypted and the "Open query" link  then I could publish this. (Deleting and being able to hide queries would be nice to have too).

mwl:
Thanks @daniellee. Your feedback is very much appreciated. We'll give it another shot.

xorsnn:
@daniellee is it true that secureJsonData can be used only for the list of rotes defined in plugin.json? How can we provide token authentication for self-hosted instances?

alexanderzobnin:
@xorsnn It seems, it's tricky to use `secureJsonData` for self-hosted instances. @daniellee any thoughts?

AlexanderBrandborg:
@daniellee Hey, this PR is rather old. Development for the plugin stalled for a while, but we just got back to it. We've now made a major upgrade to the Humio plugin  for Grafana.  Would you or someone on the team have time to review this new version? Should we in that case submit it for review in a new PR to this repo, or keep it in this thread? :- )

daniellee:
@AlexanderBrandborg it is probably better to close this and open a new PR but which ever is easiest for you.

As Humio is a commercial service, your company will need to buy a plugin subscription before I can publish this plugin: https://grafana.com/terms/

AlexanderBrandborg:
@daniellee thanks for getting back to me on this. We have decided not to proceed with becoming a featured plugin this time around. Thank you for your time :- )
You can safely close this PR for now.

PR #228, opened 2018-02-07
Add LightStep grafana plugin version 1.0.2

arya:
Blog post describing what it does:

https://lightstep.com/blog/introducing-the-lightstep-plugin-for-grafana/

### DIFF
--- a/repo.json
+++ b/repo.json
@@ -2355,6 +2355,18 @@
           "url": "https://github.com/farshidtz/linksmart-sensorthings-datasource"
         }
       ]
+    },
+    {
+      "id": "lightstep-app",
+      "type": "app",
+      "url": "https://github.com/lightstep/lightstep-grafana-plugin",
+      "versions": [
+        {
+          "version": "1.0.2",
+          "commit": "34806160577c87a4316d531ceadddb3d58e953b0",
+          "url": "https://github.com/lightstep/lightstep-grafana-plugin"
+        }
+      ]
     }
   ]
 }


arya:
@daniellee No rush, just wanted to check in to see if there is anything additional we need to do to get this PR in. Thanks!

daniellee:
Sorry for the delay - have been very busy with Grafana 5.0. Before I start reviewing, could you fix the plugin id. `grafana-lightstep-app` should either be `lightstep-app` or `lightstep-lightstep-app`.

http://docs.grafana.org/plugins/developing/code-styleguide/#plugin-json-mandatory

torkelo:
Please update the plugin ids, they need to follow a specific convention to be posted on grafana.com

arya:
@torkelo updated!

arya:
@daniellee Fixed the plugin ID in the plugin.json and updated the repo.json for this repo to point it.

torkelo:
The Lightstep app looks to include a copy / fork of the standard core Grafana Graph panel (but renamed Lightstep Graph). Is this really necessary?

daniellee:
@arya thanks for the fix. Here is my review so far:

## Plugin Review

### Metadata

- The README should link to the [docs](https://docs.lightstep.com/docs/grafana-plugin-for-lightstep)
- The plugin.json file looks fine. Though maybe a few screenshots would spice up the grafana.com page.
- The docs page does not really describe where/how to create an API key and which permissions are needed. Also, there are several settings pages - the first settings page I found (the [project settings page](https://app.lightstep.com/grafana-labs-sandbox/project)) had a list of tokens which is a similar concept to an API key. I tried using a token there before I found the Account setting page with the API keys.
- The docs don't says which API permission is needed for the Grafana plugin:
    image

### Config Page

- The config page should link to the [docs page](https://docs.lightstep.com/docs/grafana-plugin-for-lightstep).
- The labels should on the left should align vertically. You can do this by setting the same width. Grafana has some CSS classes to help with this: `width-x`
image
    Width 8 should work: `<span class="gf-form-label width-8">Project</span>`
- The error message when authentication fails could be formatted to be more understandable:
    image

- Why have the proxy/direct dropdown and the two url fields? Should the user be able to change them?

### Query Editor

- Error handling should be added ([see docs](http://docs.grafana.org/plugins/developing/plugin-review-guidelines/#query-editor)). When I get an error then no error is shown anywhere, I have to look at the console to see the error.
    image

### LightStep Graph

- It doesn't seem to work for me, can only see the legend - here is the same query in the LightStep Graph panel and in the Grafana Graph panel:

    image

### Authentication

The API key/Bearer token gets sent from the frontend which means anyone logged into Grafana can look at the HTTP request and copy it. You should use the proxy feature of Grafana. You can encrypt the API key and save it the database (using the SecureJsonData blob). If you define a plugin route for the datasource then Grafana will decrypt the api key and add it as a header to every request for you:

```json
{
  "path": "lightstep",
  "method": "GET",
  "url": "https://api.lightstep.com",
  "headers": [
    {"name": "Authorization", "content": "Bearer {{.SecureJsonData.lightstepApiKey}}"}
  ]
}
```

Documentation: http://docs.grafana.org/plugins/developing/auth-for-datasources

arya:
@torkelo 

> The Lightstep app looks to include a copy / fork of the standard core Grafana Graph panel (but renamed Lightstep Graph). Is this really necessary?

Happy to adjust the name or otherwise clarify the origin, but we copied and modified because it was the only way we could find to extend it to support the UI to display trace examples inside of a graph that were clickable.

@daniellee We're working on addressing all your feedback and will post an update soon. Thanks for the thorough review!

mattttt:
HI @arya Syncing here, do you have any updates we can review to get the plugin published?

Thanks!

cc @daniellee

bg451:
Thanks for following up @mattttt , I've included a proposal below of what is currently required in order for us to move forward on scaling this integration. Much of this may not be new information to you, but wanted to include it just in case.

LightStep [x]PM captures system-wide latency distribution data as well as distributed traces across a the spectrum of latency percentiles. Our built-in time series visualization overlays latency time series with links to example traces, in the form of clickable points, plotted at the point-in-time it occurred and the latency value

As a majority of our customers leverage Grafana for dashboarding across many backends, and have requested bringing these timeseries + trace overlays as charts into Grafana to better integrate into their workflow, we prototyped a plugin but ran into limitations with the current version of the builtin graph panel. We forked and hacked on the graph panel to support our use case, but it feels fragile and will probably break with future versions of Grafana. We've identified two features that we believe will enable this type of distributed trace access through Grafana--see below:

*Support global overrides (https://github.com/grafana/grafana/issues/10590)*
Problem: Timeseries data in the Graph Panel has a default rendering of lines. We would like trace exemplars to be graphed as dots. However, since it's not possible to set a global override user have to manually add an override to each graph.

*Request*: Allow data sources to define rendering overrides to enable rendering trace points. Particularly, for our data source, we would like to use points rather than lines by default.

*Support metadata on time series points (https://github.com/grafana/grafana/issues/10568)*
Problem: The data source API is limited to a fixed schema that is passed through to the Graph panel. This prevents us from attaching metadata to the points for the click handler to use. We want the plotted trace exemplars to be clickable and redirect to the actual trace. 

*Request*: Allow the ability to return metadata for each datapoint a datasource returns. Then take that metadata and pass it through the "graph-click" app event in the graph panel.

In order to move forward, we'd need help completing those two requests. I'll incorporate the above feedback in the meantime. Please let me know if that would be possible or if you have any other questions!

PR #231, opened 2018-02-10
Add atlas plugin

NZvereff:
[Atlas by Netflix](https://github.com/Netflix/atlas) is a backend for managing dimensional time series data.

### DIFF
--- a/repo.json
+++ b/repo.json
@@ -1967,6 +1967,18 @@
           "url": "https://github.com/devicehive/devicehive-grafana-datasource"
         }
       ]
+    },
+    {
+      "id": "upwork-atlas-datasource",
+      "type": "datasource",
+      "url": "https://github.com/upwork/grafana-atlas-datasource",
+      "versions": [
+        {
+          "version": "1.0.0",
+          "commit": "c1c70cd15d825bb5e05424ca311a839aac14f24e",
+          "url": "https://github.com/upwork/grafana-atlas-datasource"
+        }
+      ]
     }
   ]
 }


daniellee:
Sorry for the long delay here. Having trouble getting this working with Atlas 1.5.3. It seems to return invalid json. The standard json format described here: https://github.com/Netflix/atlas/wiki/Output-Formats#stdjson is not valid json as some of the values in the in the values array look like this:

```json
"values": [
  [299, 028585],

```

The `028585` value seems to fail json validation due to the leading zero. Does this work for you?

Leading zero seems to indicate that the number is octal: https://stackoverflow.com/questions/27361565/why-is-json-invalid-if-an-integer-begins-with-0

daniellee:
@NZvereff did you see my comment above?

NZvereff:
@daniellee sorry, I've missed the notification. I'll check and get back here in a week.

@briangann thank you very much. Our code was forked from your repo at some point to include these fixes.

daniellee:
@NZvereff great! My problem with the json is probably me doing something stupid. Let me know when you have checked.

NZvereff:
@daniellee , sorry for delay again. Could you please explain in more detail what exactly happened? I couldn't reproduce it locally on our live data (but I am not sure that we have any values with a leading zero there)

NZvereff:
@daniellee , sorry for delay again. Could you please explain in more detail what exactly happened? I couldn't reproduce it locally on our live data (but I am not sure that we have any values with a leading zero there)

NZvereff:
@daniellee , sorry for delay again. Could you please explain in more detail what exactly happened? I couldn't reproduce it locally on our live data (but I am not sure that we have any values with a leading zero there)

daniellee:
@NZvereff I used the Atlas with sample data from the Getting Started section in the readme: https://github.com/Netflix/atlas/wiki/Getting-Started#run-a-demo-instance

It produces numbers with leading zeros for me anyway. As that is not valid json, I get a JavaScript error in Grafana when making Ajax calls:

image

Is this just a problem with the sample data? Is there a better to test with Atlas than using the Getting Started guide?

NZvereff:
Got it, thank you. I'll check this.

NZvereff:
@daniellee sorry for delay. I have updated the plugin and checked it with the Atlas demo data.

daniellee:
@NZvereff I'm still getting the same error even for very simple queries using the sample data. Tested on Chrome and Firefox on Ubuntu.

Here are the steps I took to test the datasource with the sample data:
    ```bash
    curl -LO https://github.com/Netflix/atlas/releases/download/v1.5.3/atlas-1.5.3-standalone.jar
    java -jar atlas-1.5.3-standalone.jar
    ```

It does work if I follow the instructions in the [Running Demo with Memory Storage](https://github.com/Netflix/atlas/wiki/Getting-Started#running-demo-with-memory-storage) section.

Looking at the error, I don't see how you can fix it in the grafana datasource plugin if the Atlas sample data contains invalid data.

So assuming that it is the sample data that is incorrect, here is my review of the Atlas plugin.

## Plugin Metadata

- The readme looks ok - no broken links and shows the basics.
- It is missing a link to the github repo for the plugin - which would make it easier for users to find the repo and to open an issue.
- The rest of the metadata looks fine.

### Plugin Config Page

- Getting a JavaScript error when typing in the url field:
    ```
    bootstrap.js:1147 Uncaught TypeError: Cannot read property 'toLowerCase' of undefined
        at Typeahead.matcher (bootstrap.js:1147)
        at bootstrap.js:1134
        at Function.grep (jquery.js:424)
        at Typeahead.process (bootstrap.js:1133)
        at Typeahead.typeahead.lookup (angular-strap.js:181)
        at Typeahead.keyup (bootstrap.js:1303)
        at HTMLInputElement.proxy (jquery.js:10268)
        at HTMLInputElement.dispatch (jquery.js:5183)
        at HTMLInputElement.elemData.handle (jquery.js:4991)
    ```
- Does not suggest a url in the url field. You can do that like this (see example [here](https://github.com/grafana/grafana/blob/master/public/app/plugins/datasource/influxdb/partials/config.html#L1-L2)):

    ```html
    <datasource-http-settings current="ctrl.current" suggest-url="http://localhost:8086">
    </datasource-http-settings>
    ```
### Query Editor

image

- It does not use the Grafana CSS classes and looks a bit messy. The labels and fields should be lined up vertically. It's hard to know where to click. Functionally it all works but it is confusing. The alias field might be better at the end so that the user can choose an alias after getting a query to work. See [here](http://docs.grafana.org/plugins/developing/plugin-review-guidelines/#html-and-css) for more details on HTML and CSS for a query editor.

NZvereff:
@daniellee thank you. Looking into this.

briangann:
@NZvereff have you made any of the changes for the items that Daniel noticed?

I see another issue, the tags are not showing up, but the query to get them succeeds:

image

image

briangann:
any update from your side?

marcusolsson:
Closing due to inactivity from plugin author.

PR #232, opened 2018-02-12
Update ClickHouse datasource

hagen1778:
Here's what changed:
* properly format query with system reserved names
* columns autocompletion in ace-editor
* support array indexing int AST


daniellee:
Merged and published!

PR #233, opened 2018-02-19
Updated bubble-chart plugin.

digrich:
Removed google analytic code.


daniellee:
Merge and published. Sorry for the delay and thanks for making the change.

PR #236, opened 2018-02-27
add prometheus kubernetes-app

jtlisi:


daniellee:
Published and the old app is deprecated: https://grafana.com/plugins/raintank-kubernetes-app

PR #238, opened 2018-02-27
update d3gauge panel to 0.0.5 for grafana v5 compatibility

briangann:
This release contains a minor fix for determining panel height that was not compatible with grafana v5.

Verified this works with v3.1.1, v4.6.3, and latest v5 beta.


daniellee:
Merged and published.

PR #239, opened 2018-02-28
Plotly 0.0.4

ryantxu:
Changes:

Load plotly from npm (v1.31.2+)
Convert to TypeScript
Reasonable behavior when adding single metric
Formatting with prettier.js
Support for a single table query


daniellee:
Merged and published!

Nice work with the conversion to TypeScript. Cool that you added Prettier and also liked the copying plotly from npm in the build script. Should add those into our plugins.

PR #242, opened 2018-03-12
update gauge to v0.0.6 - vertical alignment

briangann:
BUGFIX: 
  * This update fixes a vertical alignment issue and includes updates to newer npm/bower modules.  The gauge will now auto-center vertically according to height of the panel.  Verified working on v4.6.3 and v5.0.


daniellee:
Merged and published. Nice work with the auto centering - looks great.

PR #246, opened 2018-03-15
AppDynamics version 1.2.0

dlopes7:
Better looking query-editor.html
Adds Top X feature
Adds Time Offset feature

This revamps a little bit the HTML for the query editor 

screen shot 2018-03-15 at 3 16 16 pm

It also adds support for the "Top X" feature, a key feature for when the customer has hundreds of monitoring entities and doesn't want to display them all in a single panel.

Another big feature is the "Time Offset" which lets you display metrics with a time offset on the current panel, so you can compare values from today with values from the past inside the same panel.

### DIFF
--- a/repo.json
+++ b/repo.json
@@ -577,6 +577,11 @@
       "url": "https://github.com/dlopes7/appdynamics-grafana-datasource",
       "versions": [
         {
+          "version": "1.2.0",
+          "commit": "32fecb874a563672810f864bd5ad1013d799ebd9",
+          "url": "https://github.com/dlopes7/appdynamics-grafana-datasource"
+        },
+        {
           "version": "1.1.1",
           "commit": "fdf8d2438ca44c51ec0b8af2f19f5ce226316b2e",
           "url": "https://github.com/dlopes7/appdynamics-grafana-datasource"


saurabh3896:
I can't access the git repository for your code. Has it been moved somewhere else?

dlopes7:
> I can't access the git repository for your code. Has it been moved somewhere else?

The plugin is premium so the repo is private

saurabh3896:
Got it.  Thanks. Just had a small query, can your plugin be used for querying license usage data from the app dynamics controller?

dlopes7:
No! There is no public api to get license information. 

The plugin only access the following APIs

/controller/rest/applications
/controller/rest/applications/tiers
/controller/rest/applications/{id}/tiers/{id}/node
/controller/rest/applications/{id}/metrics
/controller/rest/applications/{id}/metric-data

License information is hidden behind private APIs, I can't access any private APIs because they require a cookie to keep session and can break if the appdynamics version changes.

saurabh3896:
I see. Thank you for the information.

saurabh3896:
I was going through the source code of your plugin. Is there any way to use the 'request' module of nodeJS in your code for making GET requests?

dlopes7:
No. The requests would be made from the browser if I did that.. and show all authentication details.... furthermore AppDynamics would block Cross Origin requests.. It is not that simple, requests must be made using the Grafana backend.

saurabh3896:
I saw that GET requests are made using 'backendsrv.datasourcerequest()'. I tried to send a GET request using this method to get cookies first, store them in a variable and use it again to send another POST request to internal API for authentication. But the response I get doesn't return cookies in it's header in the first place (value is null). Though it's there in the console's Network tab. Have you ever witnessed anything like this?

dlopes7:
I'm sorry @saurabh3896 I can't give tutorials on how Grafana works here. Believe me if it was possible to use the controller /restui/ APIs I would have done already it would help me on a lot of different usecases.

You can see here that the Grafana backend will delete the Set-Cookies header.. https://github.com/grafana/grafana/blob/46ebe245abe8223d273b7b2d58278ecc176137d2/pkg/api/pluginproxy/ds_proxy.go

saurabh3896:
I see. Much thanks for your quick reply and apologies for asking too much. Great work on the plugin tho!

PR #251, opened 2018-03-22
Release ajax-panel 0.0.4

ryantxu:
LOTS of updates here.  Biggest change is that it can now be used for webcameras

Changelog
v0.0.4

- Support template variables in parameters (@linar-jether)
- Improved error handling
- Move ajax requests to 'issueQueries' block rather than refresh
- Show loading spinner
- Convert to TypeScript
- Use datasources for complex authentication
- Support loading images
- Support header configuration
- Support showing time info
- Include various sample configurations


daniellee:
There is one small thing to fix for next time - there is horizontal scroll when on the Request tab:

image

There are three divs that have `style="width: 100%" which are causing the problem.

But that is only a small problem. The examples tab is amazing and tons of new features too. Great work @ryantxu ! Merged and published.

PR #260, opened 2018-04-13
Add adremsoft-netcrunch-app 2.0

gorcz-adrem:


daniellee:
This update is fine but I noticed something that was missed in an earlier review. The password is being [saved in plain text](https://github.com/adremsoft/grafana-netcrunch-plugin/blob/master/src/datasource/config/config.html#L42). Passwords should be saved in the secureJsonData blob instead.

After changing the config page to use secureJsonData you will need a migration strategy for existing users. It should check if jsonData.password exists and if it does save it in secureJsonData.password instead and then delete jsonData.password.

gorcz-adrem:
The suggestion is valuable, but how can I get credentials from secureJsonData during datasource instance creation? NetCrunch api has own authentication method not based on HTTP basic or digest and it requires to pass credentials during establishing connection. I noticed that currently secureJsonData isn't injected into datasource instance and it's reason why jsonData has been used. Other plugins use jsonData in that same way probably for this same reason.

daniellee:
@gorcz-adrem good point - I didn't look at your api code. 

Just so you know it is possible - you can securely connect using token authentication and the Grafana proxy.

You can specify routes in your plugin.json file:

https://github.com/grafana/azure-monitor-datasource/blob/master/src/plugin.json#L30-L95

Then when you build your url in your datasource class, the url should start with the text specified in the path field for a route. The proxy will strip out the path text and replace it with the value in the url field. 

For example, if my code makes a call to url `azuremonitor/foo/bar` with this code:

```ts
this.backendSrv
      .datasourceRequest({
        url: url,
        method: 'GET',
      })
```

 then the Grafana proxy will transform it into "https://management.azure.com/foo/bar" and add CORS headers.

The token auth section looks like this:

```json
"tokenAuth": {
        "url": "https://login.microsoftonline.com/{{.JsonData.tenantId}}/oauth2/token",
        "params": {
          "grant_type":  "client_credentials",
          "client_id": "{{.JsonData.clientId}}",
          "client_secret": "{{.SecureJsonData.clientSecret}}",
          "resource": "https://management.azure.com/"
        }
      }
```

This interpolates in data from jsonData and secureJsonData. Grafana will automatically renew the token if it has expired.

You can also use an auth header:

https://github.com/grafana/azure-monitor-datasource/blob/master/src/plugin.json#L91-L93

```json
    {
      "path": "appinsights",
      "method": "GET",
      "url": "https://api.applicationinsights.io",
      "headers": [
        {"name": "X-API-Key", "content": "{{.SecureJsonData.appInsightsApiKey}}"}
      ]
    }
```

daniellee:
Merged and published!

PR #261, opened 2018-04-13
patch version to citilogics-geoloop-panel

samhatchett:
the commit hash was wrong for the v1.1.1

### DIFF
--- a/repo.json
+++ b/repo.json
@@ -2059,7 +2059,7 @@
       "url": "https://github.com/citilogics/citilogics-geoloop-panel",
       "versions": [
         {
-          "version": "1.1.1",
+          "version": "1.1.2",
           "commit": "b8aeaacdc7b5580788edee84c97216c9e3d0421a",
           "url": "https://github.com/CitiLogics/citilogics-geoloop-panel"
         }
--- a/repo.json
+++ b/repo.json
@@ -2060,7 +2060,7 @@
       "versions": [
         {
           "version": "1.1.2",
-          "commit": "ab002fa85f0031dc7c582b3b600c8e40046194fc",
+          "commit": "9700601f5e52c4964896eac702320b04ab9be121",
           "url": "https://github.com/CitiLogics/citilogics-geoloop-panel"
         }
       ]
--- a/repo.json
+++ b/repo.json
@@ -2060,7 +2060,7 @@
       "versions": [
         {
           "version": "1.1.1",
-          "commit": "5ee83aed2329aa0c0ec454325a045262ac54f7ee",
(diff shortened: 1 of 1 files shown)


daniellee:
The published version is the incorrect commit. If you want everyone to get the latest then you need to increase the version number. Otherwise, existing users will not get the latest version - plugins are only upgraded when the version number changes. 

What do you want to do? I can update v1.1.1 (only new users will get the update) or I could publish v1.1.2 so that everyone gets the change.

samhatchett:
👍 thanks, @daniellee - I didn't understand that nuance, but I got it now. appreciate your attention to detail.

daniellee:
@samhatchett you'll need to update the version number in the [plugin.json](https://github.com/CitiLogics/citilogics-geoloop-panel/blob/master/dist/plugin.json#L26) to 1.1.2 as well and then update the commit here in the plugin repo to be the latest commit.

samhatchett:
hmm. I thought i did, above. can you squash, or do you need a single clean commit from me?

daniellee:
Let me know when you have updated the version number in the plugin.json in the geoloop repo and then I can publish 1.1.2.

samhatchett:
Um, right. Now I really really understand how this works (sad trombone 🎺). I believe everything is good to go now.

daniellee:
Hi Sam,

The version is still incorrect in the [plugin.json in the dist folder](https://github.com/CitiLogics/citilogics-geoloop-panel/blob/master/dist/plugin.json#L26). (Grafana loads the dist folder if it is exists and then ignores the src folder so the plugin.json in the dist folder is the important one).

samhatchett:
thanks for sticking with me here - i think i finally got it.

daniellee:
Sorry for the delay. When I installed this and tried to add a new Geoloop panel then I got a JavaScript error:

```
Plugin component error Error: [$injector:unpr] Unknown provider: ctxSrvProvider <- ctxSrv
```
Is `ctxSrv` correct [here](https://github.com/CitiLogics/citilogics-geoloop-panel/blob/master/src/geoloop_ctrl.js#L56)? Shouldn't it be `contextSrv`?

Seems it got changed in [this commit](https://github.com/CitiLogics/citilogics-geoloop-panel/commit/9700601f5e52c4964896eac702320b04ab9be121).

daniellee:
@samhatchett did you find time to look at this?

samhatchett:
sorry @daniellee - got bogged down! I'm thinking I'll just do some more major cleanup for the next release. Closing for now. Thanks for all your guidance!

PR #263, opened 2018-04-16
Added grafana-http-json-datasource

DracoBlue:
Since simple json currently lags adhoc filter support (PR is also created https://github.com/grafana/simple-json-datasource/pull/91 ), I thought it might be a good idea to keep simple json datasource small and add extra features to an extra data source called "http json datasource".

bildschirmfoto 2018-04-17 um 00 38 26

### DIFF
--- a/repo.json
+++ b/repo.json
@@ -624,6 +624,18 @@
       ]
     },
     {
+      "id": "exozet-http-json-datasource",
+      "type": "datasource",
+      "url": "https://github.com/exozet/grafana-http-json-datasource",
+      "versions": [
+        {
+          "version": "1.4.0",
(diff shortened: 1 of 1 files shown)


daniellee:
Hi,

I merged your PR to the Simple Json datasource. I think it is a natural addition and not something that makes it bloated - especially as you can choose to support adhoc filters or not.

Do you still want to publish this plugin?

DracoBlue:
Hey @daniellee !

I think it is fine to not publish it. I will see what other features I add to http-json-datasource and will submit it for publishing again IF its necessary. For now the 1.4.0 of simple-json-datasource fits my needs! Thanks for accepting the PR.

PR #266, opened 2018-04-19
Update repo.json to add michaeldmoore-multistat-panel

michaeldmoore:
At long last, I've addressed the outstanding features to the multistat panel I tried to release some time ago. Fingers crossed, it addressed averything you had pointed out int eh earlier review - plus a lot, lot more. i think it is a pretty powerful addition, serving, especially, process monitoring dashboards.

I'm adding this to the repo.json

{
 "id": "michaeldmoore-multistat-panel",
 "type": "panel",
 "url": "https://github.com/michaeldmoore/michaeldmoore-multistat-panel",
 "versions": [
 {
 "version": "1.1.0",
 "commit": "620894eed6b1724ac6aa6babbdfef58e584cf4f7",
 "url": "https://github.com/michaeldmoore/michaeldmoore-multistat-panel"
 }
 ]
 },

### DIFF
--- a/repo.json
+++ b/repo.json
@@ -1972,6 +1972,18 @@
       ]
     },
     {
+      "id": "michaeldmoore-multistat-panel",
+      "type": "panel",
+      "url": "https://github.com/michaeldmoore/michaeldmoore-multistat-panel",
+      "versions": [
+        {
+          "version": "1.1.0",
+          "commit": "620894eed6b1724ac6aa6babbdfef58e584cf4f7",
+          "url": "https://github.com/michaeldmoore/michaeldmoore-multistat-panel"
+        }
+      ]
+    },    
+    {
       "id": "jasonlashua-prtg-datasource",
       "type": "datasource",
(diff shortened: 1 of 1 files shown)


daniellee:
Send you a PR that fixes some small things: https://github.com/michaeldmoore/michaeldmoore-multistat-panel/pull/7

The main problem that is left is that it is really easy to crash the browser if the query returns more than a few rows.

Example using InfluxDB:

1. I start writing a query and do a simple select statement against a table in InfluxDB

    image

    This returns 721 values - the time range is 6 hours and the sensor values are every 30 seconds per default:

    image

2. I change the format to Table and I have pulse/flash is chosen for the High Limit. Boom, browser hangs.

If I first change the aggregation to last and remove the group by time then it works but my first experience is that the tab crashes and I have to close it and then start again with a new dashboard.

So a query like this works:

image

Singlestat solves this with the stat option that uses the kbn library to convert multiple values into a single value.

image

daniellee:
Closing this as there is a newer PR: #352

PR #269, opened 2018-04-27
Update Breadcrumb Panel

timyllym:
Breadcrumb panel had a bug that it overwrote Grafana query-parameters


daniellee:
Merged and published!

PR #270, opened 2018-04-30
Instana datasource plugin 1.02

JeroenSoeters:
New minor version that includes 2 bugfixes

### DIFF
--- a/repo.json
+++ b/repo.json
@@ -2089,7 +2089,7 @@
         },
         {
           "version": "1.0.2",
-          "commit": "4c7f1a45708e27a0f75e04272d7d556d8a9ae7df",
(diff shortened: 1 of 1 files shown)


daniellee:
Hi,

The plugin version in the [plugin.json](https://github.com/instana/instana-grafana-datasource/blob/master/dist/plugin.json#L27) file in the dist folder is still 1.0.1. When Grafana loads a plugin, if a dist folder exists then it will load that and ignore everything else so it is important that the version number is correct there.

//Daniel

JeroenSoeters:
Fixed.

daniellee:
The change looks fine. Merged and published.

I have some thoughts about the datasource config page that I will send in an email to you.

PR #271, opened 2018-05-02
 Zabbix plugin version 3.9.1

alexanderzobnin:
Bug fixes introduced by Grafana 5.1 release.


daniellee:
Published - me and Marcus will have a look at MySQL ds and why it cannot handle numbers as metric names.

PR #280, opened 2018-06-02
Adding Pagerduty datasource

xginn8:


daniellee:
Quick note on why the build is failing - the plugin id in this PR does not match the plugin id in the [src/plugin.json file](https://github.com/xginn8/grafana-pagerduty/blob/master/src/plugin.json#L3) and the [dist/plugin.json file](https://github.com/xginn8/grafana-pagerduty/blob/master/dist/plugin.json#L3).

xginn8:
@daniellee thanks for that protip! PTAL when you can!

daniellee:
Here is my feedback on the PagerDuty Datasource.

## Metadata Feedback

The plugin.json looks good.

### Readme

The Readme is missing a few things:

- Should say more clearly that this is an Annotation only datasource
- Should describe or link to a description of how to create an API key
- Should explain what a Service ID is and where to find it

## Plugin Feedback

### Config Page

- The API key is not encrypted. See the [Config Page Guidelines](http://docs.grafana.org/plugins/developing/plugin-review-guidelines/#config-page-guidelines) and the [Authentication for Datasources](http://docs.grafana.org/plugins/developing/auth-for-datasources/) docs for more details.
- No error message if the API key is incorrect:
    image

## Annotation Query Editor

- No description or help for the three filter fields.

## Plugin

It doesn't work for me. Getting an error on this line: https://github.com/xginn8/grafana-pagerduty/blob/master/src/datasource.js#L58 

`SyntaxError: Unexpected token u in JSON at position 0`

Looking at the code, I don't see any field called `query` in the `annotation` object and it looks like the query variable is never used.

daniellee:
@xginn8 how is it going with this?

xginn8:
@daniellee thanks for following up! I've been preoccupied lately but will hopefully have something in the next week or two.

xginn8:
@daniellee updated finally, PTAL!

daniellee:
Sorry for the long delay. Tested this out and it works now but it has a few areas of improvement for the next version:

- The drop-downs are overlapping the info popovers:

image

- The readme contains a link to this page to create an api key: https://support.pagerduty.com/docs/services-and-integrations but it looks like that is not the right page? I think it is this one: https://support.pagerduty.com/docs/using-the-api

Alex published it so thanks for your patience!

PR #282, opened 2018-06-07
Added traffic-light plugin

vmanilo:


daniellee:
Could you make this image a bit smaller: https://github.com/smartmakers/grafana-trafficlight/blob/master/dist/img/view.png

Either with Image Magick - something like this:

`mogrify -resize 80% src/img/view.png` 

or some other tool like [pnggauntlet](https://pnggauntlet.com).

vmanilo:
Hey, @daniellee, thanks for suggestion, I downsized this image

vmanilo:
Do I need to change commit hash in repo.json?

daniellee:
@vmanilo yes, please.

vmanilo:
@daniellee done

daniellee:
Merged and published: https://grafana.com/plugins/smartmakers-trafficlight-panel

## Plugin Feedback

Everything works as it should and the readme looks fine. 

There is one thing that could be improved and that is that the values returned from the query must be 1, 2 or 3. Some sort of threshold management would make this plugin more accessible so that a user could assign a range of values to be red/amber/green -> 0-100 == red, 101-200 == amber,  201+ == green

We are redoing the threshold function in the singlestat and other similar panels so we could export that as a lib that you and other plugin authors could use. That will be released in August or September.

vmanilo:
@daniellee I appreciate your help 👍

PR #286, opened 2018-06-22
Release prometheus-alertmanager-datasource 0.0.6

cryptobioz:


daniellee:
Merged and published. Interesting to see that you switched to the apache license. And we should definitely fix that provisioning bug in Grafana.

PR #287, opened 2018-06-23
Submit a new panel that make windrose plot

fatcloud:
the first working version

### DIFF
--- a/repo.json
+++ b/repo.json
@@ -2217,7 +2217,7 @@
       "versions": [
         {
           "version": "0.7.0",
-          "commit": "121754fa12dbe67050e0e20dd12a53811f313a18",
+          "commit": "894dc65f2c0188120f5c9ee1345257f675d626eb",
           "url": "https://github.com/fatcloud/windrose-panel"
         }
       ]
--- a/repo.json
+++ b/repo.json
@@ -2217,7 +2217,7 @@
       "versions": [
         {
           "version": "0.7.0",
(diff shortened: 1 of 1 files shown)


daniellee:
Before I start reviewing this plugin, can you fix the paths to the screenshots in the plugin.json, please. Grafana tries to load the `dist` subdirectory if it exists and the paths to the screenshots are relative to the plugin.json in the dist folder. 

I recommend moving the plugin.json file into the `src` subdirectory - this makes the paths for the images simpler.

The first screenshot [path](https://github.com/fatcloud/windrose-panel/blob/master/plugin.json#L19) in the plugin.json file is `src/img/twocharts.png`. If you move the plugin.json file into src then it can be `img/twocharts.png`. This will work both for `src` and `dist`.

fatcloud:
Done moving plugin.json and updating dist directory.

Actually plugin.json WAS placed in src in my first three commits in this PR.
But I couldn't pass the auto test somehow, and I moved it to top level in desperation by commit 6206eed,
which magically worked out.
So it's rather confusing for me to see this modification didn't break auto test again. lol

daniellee:
Testing this out and it looks good. The only small issue is that if there is an error contacting the datasource then I have to refresh the page to get it work afterwards:

image

But you can fix that in the next version - this looks good to publish! I can fix the merge conflict.

daniellee:
Merged and published: https://grafana.com/plugins/fatcloud-windrose-panel

PR #289, opened 2018-07-03
OpenNMS Helm 2.0.0

j-white:
Updating the OpenNMS Helm plugin to v2.0.0.

This update includes minor bug fixes and a new datasource for querying flow datasource from OpenNMS.

### DIFF
--- a/repo.json
+++ b/repo.json
@@ -649,6 +649,11 @@
       "url": "https://github.com/OpenNMS/opennms-helm",
       "versions": [
         {
+          "version": "2.0.0",
+          "commit": "1a6f347e19924e9c4e6e02b38b26a7dbbba95e2b",
+          "url": "https://github.com/OpenNMS/opennms-helm"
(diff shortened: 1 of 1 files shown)


daniellee:
Is the opennms demo site down?

https://demo.opennms.org is not working for me anymore.

daniellee:
I looked through the changes and tested without access to a datasource and it looks fine. Though the error message for the new Flow datasource when testing and saving is not as good as the other two.

Performance DS:

image

Flow DS:

image

To test properly, should I setup an OpenNMS environment with docker or will you be able to get the Demo site working again?

j-white:
@daniellee the demo site should be back online now.

Note that the demo server doesn't have any flow data currently, so you won't be able to get much out of the Flow DS.

daniellee:
Merged and published.

j-white:
Thanks @daniellee!

PR #293, opened 2018-07-27
Instana datasource plugin 1.1.0

basti1302:
Adds a checkbox that let's user enable the new Instana application model, yielding new/additional entity types and metrics.


basti1302:
Hi @marefr,

this was indeed a problem cause by the Mountebank config being slightly outdated. This is fixed now and I have updated this PR with the correct commit hash. Also, the procedure to test this needs to be updated a little, here is a step-by-step guide:

* `cd instana-datasource`
* `docker-compose up mountebank`
* Create a datasource for Instana in Grafana
    * URL: `http://localhost:8010`
    * API Token: `valid-api-token`
* Create a new dashboard with a graph panel
* Query: `filler`
* Type: `Processes`
* Metric: `Virtual (mem.virtual)`

This should render a chart with two datasets (`node (on host "host-1")` and `node (on host "host-2)`).

marefr:
Thanks. Followed your steps and got it working, however after saving dashboard going back and forth between home and dashboard the graph sometimes load and sometimes not. Going into edit mode the metric selection is gone each time, see screenshot.

image

Not sure if this is a test setup issue or real issue, please have a look.

basti1302:
Fixed.

marefr:
Published and merged. Thanks

basti1302:
Thanks!

PR #294, opened 2018-08-02
Add new panel: gretamosa-topology-panel

gretamosa:
Sigma.js topology panel for Elasticsearch and InfluxDB.

### DIFF
--- a/repo.json
+++ b/repo.json
@@ -2309,6 +2309,18 @@
           "url": "https://github.com/GoshPosh/grafana-meta-queries"
         }
       ]
-    }
+    },
+    {
+      "id": "gretamosa-topology-panel",
+      "type": "panel",
+      "url": "https://github.com/gretamosa/gretamosa-topology-panel",
+      "versions": [
+        {
+          "version": "1.0.0",
+          "commit": "a632d538e5a71ff339d0ac2aeba805d3f3df2a1d",
+          "url": "https://github.com/gretamosa/gretamosa-topology-panel"
+        }
+      ]
+    }	  
   ]
 }


daniellee:
Hi,

Before I start reviewing, there are a few things that need to be fixed first. 

- Grafana has no way of compiling/transpiling from ES6 Javascript to plain JavaScript so the `dist` subdirectory needs to be committed.
- The urls for the images in the README need to be changed. GitHub does lots of clever things to make sure any image link works. Those links are not to the raw image and get this error in Grafana as they are html pages and not images:
    image
    See the [plugin guidelines section on README's](http://docs.grafana.org/plugins/developing/plugin-review-guidelines/#readme-md) for an example of how the url should be.
- Links to some css files look incorrect (should be plugin id and not `topology-panel`): https://github.com/gretamosa/gretamosa-topology-panel/blob/master/src/module.js#L5-L6
- Link to editor partial looks incorrect as well: https://github.com/gretamosa/gretamosa-topology-panel/blob/master/src/topology_ctrl.js#L56
    image

Very curious to see what you've done with Sigma.js!

gretamosa:
Hi Daniel,

Apologies for submitting this pull request without verifying these issues. In last minute, I changed the name of the project, from topology-panel to gretamosa-topology-panel and I forgot a "replace all" command.

I think I've resolved all pending issues (update repo.json with the latest commit change). Tell me if there is something more wrong ;)

daniellee:
@gretamosa do you have an example of an InfluxDB query or an Elasticsearch query that works with this panel?

gretamosa:
In fact, you can use a wildcard query (Elasticsearch) or InfluxDB you can use default or another measurement with fields that your are going to use in next steps to make it work.

`SELECT "callerip", "calleeip", "calleeloss", "calleeloss", "calleemos", "callermos" FROM "rtp" WHERE $timeFilter GROUP BY "type")`

Then, you must to go to "Options" tab and select "Source Field" as your source node label and "Destination Field" as your destination node label. Finally, if you want to show edges with different widths according to a specific value complete "Edge Size Field".

daniellee:
Hi @gretamosa 

Having problems understanding what sort of queries work with this panel. Your example query looks invalid to me, it has a group by but no aggregations for the fields.

Can you give a typical use-case and explain how and why you would use the topology panel. I can get a query to work if it has no group bys  - like a log table. But I'm not sure why that would be useful - I was thinking I would be able to use it like the Diagram panel and use it to create a topology with the edges in between showing the flow of data.

Basically, what I am trying to say is that this plugin needs documentation. It is hard to understand what the different options are and how they match up to fields returned by the query.

gretamosa:
Hi @daniellee 

I understand your problems. The use-case scenario that I am currently using is with network topologies where I have source IP address (source field in Options > Connections section), destination IP address (destination field in Options > Connections section).

We use Grafana in multiple IT clients with a number of networking use-cases and by establishing edge size field according to specific metrics (i.e. RTT, incoming/outgoing bytes, CET...) provides useful panels because without it I can assure you that clients understand this kind of panels like a toy.

Finally, grouping is not based in data context but in sigmaJS graph context. If you are familiar with sigmaJS, nodes are randomly placed but you can implement grouping node algorithms, according to their neighboorhood and some other criteria, to have a visible panel (not just crossing edges between nodes). Back to my use-case, it could be some other network parameters such as VLAN identifier, protocol, geographical locations.

I am agree with you that documentation could be improved (this is a first approach) but it is hard to establish a global documentation. It is an improved use of Diagram panel because in a diagram panel you cannot customize edge types (so important in a lot of use-cases), you cannot group nodes (almost more important than previous one) and, as i remember from the last time that I used this panel, you cannot set an edge size field and the most important, it is not necessary to learn and maintain mermaidJS :)

Hope it helps and I am going to try to improve my documentation to be more descriptive and useful.

pbuxtonccuk:
Not sure if this is the right place, but when I install this panel, I can't actually use it.   I see a Panel Plugin Not Found message when I try to add one.

gretamosa:
Follow instructions in https://github.com/gretamosa/gretamosa-topology-panel. It's the better way and the proper place to report issues.

PR #295, opened 2018-08-07
 Update clickhouse-datasource to 1.6.0

hagen1778:
* annotations support (txh to @atsirin)
* allow to use `$from` and `$to` macroses in variable queries
* provisioning config example in README


daniellee:
Nice that you added support for annotations - works well.

Merged and published.

PR #299, opened 2018-08-21
Update to Grafana-breadcrumb-panel

timyllym:


daniellee:
Wrong version in the [plugin.json in dist](https://github.com/digiapulssi/grafana-breadcrumb-panel/blob/master/dist/plugin.json#L12). If you can update that to 1.1.4 then I can publish tomorrow.

timyllym:
Version is now updated

torkelo:
published!

PR #179, opened 2017-10-11
Adding grafana-warp10-datasource plugin

LostInBrittany:
Adding Grafana  Warp 10 Datasource Plugin, a plugin that allows Grafana 4 to support [Warp 10](http://www.warp10.io/) as datasource.

PR #302, opened 2018-09-04
feat(warp10): New plugin

miton18:
The Warp 10 Platform is designed to collect, store and manipulate sensor data. Sensor data are ingested as sequences of measurements (also called time series). The Warp 10 Platform offers the possibility for each measurement to also have spatial metadata specifying the geographic coordinates and/or the elevation of the sensor at the time of the reading. Those augmented measurements form what we call Geo Time Series® (GTS).

Let's add [Warp10.io](http://www.warp10.io/) as new data source.

### DIFF
--- a/repo.json
+++ b/repo.json
@@ -2446,6 +2446,16 @@
           "version": "0.0.1",
           "commit": "cf31fded9d3cd43ab63092064beef4a45123cf69",
           "url": "https://github.com/flant/grafana-statusmap"
+    },
+    {
+      "id": "ovh-warp10-datasource",
+      "type": "datasource",
+      "url": "https://github.com/ovh/ovh-warp10-datasource",
+      "versions": [
+        {
+          "version": "2.1.1",
+          "commit": "3f1ec1ae81fc86d5bfa0b52e98119beaee42609a",
+          "url": "https://github.com/ovh/ovh-warp10-datasource/pull/26"
         }
       ]
     }


daniellee:
Hi, can I close this #179 ? Guess that you have made lots of changes since that PR was submitted?

miton18:
@daniellee yes, you can close it :-)

miton18:
Is it OK for you ?

daniellee:
So everything works and I will publish this but I think the query editor could be improved.

- It is strange to have a read token in the query editor - means every user has to have a token rather than an admin setting up the datasource once and then working for everyone else.
- No intellisense/autocomplete for any fields (or dropdowns with values)
- Would it be possible to have an oauth endpoint for authentication? Grafana has [built-in support for that](http://docs.grafana.org/plugins/developing/auth-for-datasources/) that would mean the user would not need to know about read tokens as everything would be configured on the config page.
- Spelling mistakes: `Select on wich label, the filter must me applied` - wich and me

daniellee:
I was just about to merge when I saw that the plugin id is incorrect.

> The convention for the plugin id is [github username/org]-[plugin name]-[datasource|app|panel] and it has to be unique. Although if org and plugin name are the same then [plugin name]-[datasource|app|panel] is also valid. The org cannot be grafana unless it is a plugin created by the Grafana core team.

So in your case, it should probably be `warp10-datasource` or `ovh-warp10-datasource`. It should be changed in this PR, as well as in the src/plugin.json and dist/plugin.json files.

miton18:
@daniellee Repository renamed and PR rebased

daniellee:
Finally! Merged and published: https://grafana.com/plugins/ovh-warp10-datasource

PR #303, opened 2018-09-05
Update clickhouse-datasource to 1.7.0

hagen1778:
* provide $adhoc macros for using ad-hoc filters in inner queries (thx to @vavrusa)
* allow to set custom query for ad-hoc filter via adhoc_query_filter variable
* provide new Round value $step for auto-rounding according to graph resolution changes


torkelo:
published!

PR #304, opened 2018-09-07
Add new panel that shows user annotations

ryantxu:
This is a simple panel to show the user annotations, as discussed here:

https://github.com/grafana/grafana/issues/11323


daniellee:
Merged and published: https://grafana.com/plugins/ryantxu-annolist-panel

PR #305, opened 2018-09-08
added consul-datasource plugin v0.1.5

sbueringer:
Hi,

I'm not sure what an iditomatic plugin-id would be, so I just chose `consul-datasource`. Test environment should be more or less `npm run testcoverage` and `npm run go-testcoverage` (there is also a .travis.yml file).

Please let me now if I missed something ;)

### DIFF
--- a/repo.json
+++ b/repo.json
@@ -2355,6 +2355,18 @@
           "url": "https://github.com/farshidtz/linksmart-sensorthings-datasource"
         }
       ]
+    },
+    {
+      "id": "sbueringer-consul-datasource",
+      "type": "datasource",
+      "url": "https://github.com/sbueringer/grafana-consul-datasource",
+      "versions": [
+        {
+          "version": "0.1.5",
+          "commit": "60f73d5aef6a64068379e7625b99d0d8114b8ee1",
+          "url": "https://github.com/sbueringer/grafana-consul-datasource"
+        }
+      ]
     }
   ]
 }


sbueringer:
So `sbueringer-consul-datasource` would be a correct plugin-id right?

sbueringer:
I changed the id to `sbueringer-consul-datasource`

daniellee:
Thanks for submitting this. Here's my feedback so far:

- The config page could be improved with some instructions on which form the url should be in (or a link to instructions) and also how to create a token.
- The epoch that is returned in the time series data is not in the right format. It works in Singlestats and Tables but not in the Graph panel.

    The time field looks like this: `1538261136661525800` which is in microseconds. The graph panel expects millisecond precision e.g. 1538261269000

- Is it possible to get any sort of intellisense/autocomplete for the key values?

- This plugin is using an alpha feature of Grafana - backend plugins. The backend plugin works fine. The problem is with the distribution of the plugin which we have not worked on or solved yet. For example, when I installed this plugin I had to set the permissions of the `grafana-consul-plugin_linux_amd64` manually to get it to work. Not sure what to do about this - will discuss it with the team.

sbueringer:
Thanks for the review!

* I updated the config page with some additional instructions. 
* I fixed the precision of the timestamps to milliseconds. 
* Where are you missing the autocomplete? It should be there in the query editor
* I'm not sure if it helps during the installation of the plugin, but I've changed the build so the go binary is executable after the build.

daniellee:
Thanks for the fixes. Will get back to you next week.

sbueringer:
@daniellee Regrading the executable bit of the backend plugin:
Is this already fixed in some newer Grafana version or is there an issue in the Grafana repositories which I can subscribe to?

PR #306, opened 2018-09-14
Release usgs-datasouce 0.0.2

ryantxu:
Update the datasource to 0.0.2, The most important change is that USGS moved their service URL and this now avoids the 301 error.

This now nicely picks the mean daily value when you zoom out the instantaneous service to a large time


torkelo:
published to grafana.com!

PR #307, opened 2018-09-18
Add flant-statusmap-panel 0.0.1

diafour:
New plugin to visualize changes of statuses of multiple objects over time.


marefr:
We seem to have some problems with our CI builds. Will trigger rebuild as soon as we fixed this.

diafour:
Yeah, mapped colors got more attention than gradient. Thank you for approval.

PR #309, opened 2018-09-21
Update Akumuli datasource

Lazin:
- Support interval template variables.
- Fix safety limit.


marefr:
We seem to have some problems with our CI builds. Will trigger rebuild as soon as we fixed this.

PR #310, opened 2018-09-25
yesoreyeram/yesoreyeram-boomtable-panel v0.4.0

yesoreyeram:
* Time based thresholds (  [Details](https://github.com/yesoreyeram/yesoreyeram-boomtable-panel#time-based-thresholds) )
* Option to modify row/column identification wrapper 
* Option to reorder exsiting patterns 
* Minor bug fixes

### DIFF
--- a/repo.json
+++ b/repo.json
@@ -2289,12 +2289,12 @@
             "commit": "353d36d90a198a2251aa932a59f7589f22850d71",
             "url": "https://github.com/yesoreyeram/yesoreyeram-boomtable-panel"
           },
-	  {
+          {
             "version": "0.3.0",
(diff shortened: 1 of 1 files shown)


yesoreyeram:
Thanks @daniellee for the detailed review. Will fix the issues you mentioned and update the pull request. 

For the first issue you mentioned (Confusing UI), Does renaming the tab to "Time based thresholds" from "Thresholds" will make sense rather than moving control between the tabs? Also I will make the threshold value in the new tab to read-only. So there wont be confusion. (Actually both controls referrening the same ng-model). Looking forward for your suggestins.

PR #311, opened 2018-09-28
add thruk datasource

sni:
This datasource supports thruks (https://thruk.org) rest api which allows
various table requests for nagios/naemon/icinga setups.

Currently it supports annotations to add log events like notifications to
existing graphs.

It supports variables to use hostnames/hostgroups and stuff like that in
existing dashboards.

It also supports table panel queries to show any data from the rest api which
includes alerts, current status and such things.

In order to test this datasource, you can use https://demo.thruk.org/thruk/ with basic
auth enabled using admin/admin credentials.

### DIFF
--- a/repo.json
+++ b/repo.json
@@ -1865,6 +1865,18 @@
       ]
     },
     {
+      "id": "sni-thruk-datasource",
+      "type": "datasource",
+      "url": "https://github.com/sni/grafana-thruk-datasource",
+      "versions": [
+        {
+          "version": "1.0.1",
+          "commit": "050d770946ad0f12750d206055b80d9389e9ea52",
+          "url": "https://github.com/sni/grafana-thruk-datasource"
+        }
+      ]
+    },
+    {
       "id": "digiapulssi-organisations-panel",
       "type": "panel",
       "url": "https://github.com/digiapulssi/grafana-organisations-panel",


daniellee:
Thank you for providing a test environment - makes reviewing a datasource plugin so much easier!

I have a couple of questions:

- Just to be certain - this datasource only support table queries and not time series queries. Is that correct?

- I tested the annotation query (from the docs):

    `SELECT time, message FROM logs WHERE time = $time and host_name= 'linx.net'`

    But this produces an error:

    ```
    AnnotationSrv.query error Error: query syntax error, first column must be 'time' for annotations.
        at ThrukDatasource.annotationQuery (:3000/public/plugins/sni-thruk-datasource/datasource.js:82)
        at :3000/public/build/2.2916d419121f1c17764f.js:1904
        at processQueue (app.2916d419121f1c17764f.js:20853)
        at app.2916d419121f1c17764f.js:20897
        at Scope.$digest (app.2916d419121f1c17764f.js:22035)
        at ChildScope.$apply (app.2916d419121f1c17764f.js:22333)
        at HTMLButtonElement.<anonymous> (app.2916d419121f1c17764f.js:31148)
        at HTMLButtonElement.dispatch (app.2916d419121f1c17764f.js:80929)
        at HTMLButtonElement.elemData.handle (app.2916d419121f1c17764f.js:80737)
    ```

    And when I debugged the JS, I see that on this [line](https://github.com/sni/grafana-thruk-datasource/blob/master/src/datasource.js#L43) that query.columns is not an array of column names but a string 'time, message'. Am I doing something wrong or is this a bug?

sni:
Oh, my bad. Thats indeed a bug caused by some last minute changes :-) It's fixed now and i updated and rebased this PR.

And yes, this datasource only supports table queries, no time series. A typical setup would consist of Thruk in combination with Naemon,Icinga or Nagios to fetch metrics and Influxdb for storage. So you would create dashboards with those influx metrics and enrich it with annotations and table data from this datasource.

daniellee:
Great - thanks for the fix. For the next version, could you add to the readme that it does not support time series queries (you could take what you wrote above). It says it supports table queries and annotations but doesn't explicitly say that it doesn't support time series queries. But that is only nitpicking really. Nice work!

There is a problem with the logo but as it works locally, I'm guessing the problem is on our side. 

Anyway, merged and published: https://grafana.com/plugins/sni-thruk-datasource

and will get back to you about the logo as soon as I figure out why it didn't get uploaded properly.

PR #312, opened 2018-10-03
Update akumuli-datasource

Lazin:
Bugfix. Alias can be used with TopN query now.


daniellee:
Merged and published.

PR #313, opened 2018-10-03
updating ambari metrics datasource to v1.2.0

prajwalrao:
- Grafana > v4.5.x support added
- fixed [#8](https://github.com/prajwalrao/ambari-metrics-grafana/issues/8)

### DIFF
--- a/repo.json
+++ b/repo.json
@@ -1028,6 +1028,11 @@
       "url": "https://github.com/prajwalrao/ambari-metrics-grafana",
       "versions": [
         {
+          "version": "1.2.0",
+          "commit": "d429f53128fc8db3be7f719f153d228b643189eb",
+          "url": "https://github.com/prajwalrao/ambari-metrics-grafana"
+        },
+        {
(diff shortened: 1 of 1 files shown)


daniellee:
Great to see that this is still alive! 

I don't think that this plugin was ever tested properly by us - is there a test environment that I could use to test this? Or should I use Vagrant: https://cwiki.apache.org/confluence/display/AMBARI/Quick+Start+Guide

prajwalrao:
Thanks @daniellee  
I use Vagrant myself to spawn up a cluster(locally) and do all testing and development. I'm not aware of any test environment that's publicly available.

prajwalrao:
@daniellee Thank you so much for persevering and managing to test it. I will implement the feedback and get back to you shortly. :)

daniellee:
I should add that otherwise it works well and that these are only small issues. I think changing the readme and config page to mention the port number would be very helpful to new users. If you can fix that then I can publish this straight away.

PR #315, opened 2018-10-05
Update BT trend-box and alarm-box panels

jonasbleyl:
Fixes BT alarm-box and trend-box panels overflowing in Grafana 5.

### DIFF
--- a/repo.json
+++ b/repo.json
@@ -44,6 +44,11 @@
           "version": "0.1.8",
           "commit": "ab4ce90dcf56f7cc9e9a0d67d00a939746c1d687",
           "url": "https://github.com/BT-OpenSource/bt-grafana-trend-box"
+        },
+        {
+          "version": "0.1.9",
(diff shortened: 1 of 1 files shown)


daniellee:
Great to see an update! It's been a while. The change is merged and published.

Out of curiousity, may I ask why you set the properties in JavaScript instead of with CSS? Is it something to do with the loading in Grafana?

jonasbleyl:
Thanks @daniellee!

That CSS is overriding some of Grafana's default behaviour for the panel. 
E.g. Our panel let's you choose the Panel Title size. Unfortunately when you set this it now also changes the font size of Grafana's dropdown-menu for a panel.

When trying to fix this using CSS it applied it to every panel on a dashboard whereas with the JavaScript solution it only applied it ours. If there's a way to only override Grafana's CSS for our Panel then that would be a better solution since the overrides don't need to be dynamic. I couldn't see how to do that though unless I was missing something :)

PR #317, opened 2018-10-12
Update repo.json to add scadavis-synoptic-panel

riclolsen:
Please review my plugin: scadavis-synoptic-panel.
A powerful SCADA-like synoptic graphics panel for Grafana.
Free-form vector graphics can be created in SVG format using an Inkscape based editor.

### DIFF
--- a/repo.json
+++ b/repo.json
@@ -2438,6 +2438,18 @@
           "url": "https://github.com/flant/grafana-statusmap"
         }
       ]
+    },
+    {
+      "id": "scadavis-synoptic-panel",
+      "type": "panel",
+      "url": "https://github.com/riclolsen/scadavis-synoptic-panel",
+      "versions": [
+        {
+          "version": "1.0.0",
+          "commit": "aa5d44dc9c9ce4be4c65ec5578fd1989f4218b6c",
+          "url": "https://github.com/riclolsen/scadavis-synoptic-panel"
+        }
+      ]
     }
   ]
 }
--- a/repo.json
+++ b/repo.json
@@ -2445,8 +2445,8 @@
       "url": "https://github.com/riclolsen/scadavis-synoptic-panel",
       "versions": [
         {
-          "version": "1.0.1",
-          "commit": "46950eed6cafd7b6220b0c13144b7fa79acf6b68",
+          "version": "1.0.2",
+          "commit": "9ad761917abff9ec43a2b90ad47939fbc16c5c01",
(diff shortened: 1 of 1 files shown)


daniellee:
Hi @riclolsen 

Great to see a SCADA diagram integration with Grafana. Here's my feedback so far:

## Metadata

- There is an error in the plugin.json file. It refers to a non-existent image: https://github.com/riclolsen/scadavis-synoptic-panel/blob/master/src/plugin.json#L23
- The paths for images are relative to the location of the plugin.json file. You have two img directories in dist `img` and `src/img` but that is unnecessary. Instead of `src/img/scadavis-power.png`, change to `img/scadavis-power.png`.
- The README.md looks good.

## Plugin

- On the options tab, the labels should have the same width so that they line up:

    image

- Is it possible to center the image or set X and Y coordinates? For me, some of the image was outside of the panel. I could move it with the pan controls but it reverted to being partially off-screen after refreshing the page.

- The rendering of the legend could use some improvement. The scadavis logo ended up in various strange places depending on the legend options:

    image

- Is it possible to show what the tag/label names on the image are - to be able to match them with the query? Or is it only possible to see that inside the scadavis editor?

riclolsen:
Thanks, @daniellee,
I have corrected the problems you pointed.
The plugin shows the linked tags when mousing over them (when values are shown as numbers).
About the positioning of the image:
The plugin uses an SVG file, it grows the viewport to occupy all the available space of the panel (the image is seen as a big canvas). This way the drawing inside the SVG canvas is always at the top left positions, the zoom can be adjusted.
I have hidden the logo watermark to lessen the problem.
Please review the new version.

daniellee:
Thanks for making the changes. This is good enough to publish now. I have one question/request first though. Can you write a sentence or two more about matching tags to metric aliases in the readme. 

When I mouse over the panel, I see the following pop up:

image

Are those the tags? How do I match them to a query? What should the alias look like to match one of the tags?

An example query would be help a lot.

riclolsen:
Thanks for the meaningful suggestions, @daniellee.
I have updated the README.md to include a simple step-by-step example.
Also, fixed the mouse over function that shows tags as tooltips. It was unable to show the tag names in some situations.
Please review the 1.0.2 version.

daniellee:
Looks good - I can fix the merge conflict and publish.

daniellee:
Published: https://grafana.com/plugins/scadavis-synoptic-panel

PR #319, opened 2018-10-23
Instana datasource plugin 1.2.1

schmidtl:
- extend built-in metric catalog
- remove app 1.0 support flag


daniellee:
Merged and published.

PR #320, opened 2018-10-23
Add enhanced JSON datasource

andig:
This is a type-script written fork of the original SimpleJSON datasource. 

Most notably, it includes the ability to pass additional parameters as part of the `/query` request to allow differentiated responses by backend adapters:

screen shot 2018-10-23 at 19 50 01

One example backend adapter is https://github.com/andig/gravo which accepts grafana json queries and extracts multiple metrics simultaneously from an underlying [volkszaehler](https://volkszaehler.org) REST api. Additional JSON parameters are used to provide fine-grained control on how to use the REST api (in the screenshot above by retrieving pre-aggregated data). In that respect it provides similar functionality, though with less abstraction due to its generic nature, as the Prometheus datasource.

Upstream PRs for https://github.com/grafana/simple-json-datasource/pull/100, https://github.com/grafana/simple-json-datasource/pull/103, https://github.com/grafana/simple-json-datasource/pull/104 have been applied.

### DIFF
--- a/repo.json
+++ b/repo.json
@@ -2433,6 +2433,18 @@
       ]
     },
     {
+      "id": "grafana-json-datasource",
+      "type": "datasource",
+      "url": "https://github.com/simPod/grafana-json-datasource",
+      "versions": [
+        {
+          "version": "0.1.0",
+          "commit": "8919f9b76e07fd6b7bf3f6c6b4682cb48bdeb77d",
+          "url": "https://github.com/simPod/grafana-json-datasource"
+        }
+      ]
+    },
+    {
       "id": "flant-statusmap-panel",
       "type": "panel",
       "url": "https://github.com/flant/grafana-statusmap",
--- a/repo.json
+++ b/repo.json
@@ -2433,13 +2433,13 @@
       ]
     },
     {
-      "id": "grafana-json-datasource",
+      "id": "simpod-json-datasource",
       "type": "datasource",
       "url": "https://github.com/simPod/grafana-json-datasource",
       "versions": [
         {
           "version": "0.1.0",
-          "commit": "8919f9b76e07fd6b7bf3f6c6b4682cb48bdeb77d",
+          "commit": "c2b80cf2e9d57060968e71b7b2001c7fc816b03f",
           "url": "https://github.com/simPod/grafana-json-datasource"
         }
       ]


andig:
Removed wip label, asking for review and merge :)

zak-b2c2:
I am definitely a big supporter of this!

torkelo:
Not sure we want to publish a duplicate of the simple-json data source. The refactor to typescript would be a great PR to the original repo. I know this will sound pretty terrible as we have been so unresponsive on PRs and issues for the original simple-json-datasource. I think if we refactor it to typescript we could potentially move it to a built-in core data source (and write a go backend part would make it support alerting).

simPod:
Ummm...

...

...

@andig summed it up very well but in short this is what happened:

1. We try to contribute to original repo with no response from maintainers for almost two months even though we tried to ping using metions. Thus we consider the repo abandoned/not worth Grafana team's effort (so far it's fine, happens).
2. We start working on a fork.
3. After some time, we receive message from Grafana that we should probably fork, that the original datasource is rather meant as a template for more feature-rich forked versions and **should stay as lean as possible**. And that if we build that fork it can be pushlished.
4. We put some effort into it then, rewrite to typescript, add some features so it can do things that are essential for users of APIs.
5. We try to apply for publishing.
6. Grafana says they don't want to publish it but rather **extend the original plugin**.

I'm not sure where to go from here. This looks like Grafana states the exact opposite each month :/

briangann:
The easiest way forward is to publish this as a new plugin.

You would need to change the plugin "id" in src/plugin.json to "id": "simpod-json-datasource" to differentiate it from the original plugin (name conflict).

We can plan out how to integrate this version into Grafana core afterwards.

If you make these changes I'll work to push through the review and get it published.

Thanks!

Brian

andig:
Thanks @briangann. The original id is grafana-simple-json-datasource, we have grafana-simple-json-datasource, so no conflict. Do you still need this change?

I‘d also appreciate any suggestions how we can contribute making json a backend plugin AND allowing our requirements to run on top if not in the core.

briangann:
Our convention for plugin id is githubusername-awesomeproject-plugintype, updating the id to `simpod-json-datasource` would work best (or even simpod-ejson-datasource).  The github repo name doesn't have to change, just the id.

(you can also give it a different icon/logo so it stands out as different, but not mandatory)

I'll spin up the plugin and see what it would take for backend plugin integration.  The process isn't documented well, but it is possible to create one and included it with a non-core plugin.

repo.json
```diff
@@ -2432,6 +2432,18 @@
         }
       ]
     },
+    {
+      "id": "grafana-json-datasource",
```
briangann: this will need to be updated to match the new id also

andig:
Yes, will do once we have another fix in.

andig:
@briangann update- hope this works now.

andig:
Great, thank you :)

briangann:
you can see the plugin here: https://grafana.com/plugins/simpod-json-datasource

also available with grafana-cli

PR #321, opened 2018-10-23
Added Diagram v1.4.5

jdbranham:


jdbranham:
I've actually been considering moving the development under my business github account @Savantly-net, rather than my personal.
I'm assuming I can just update the plugin repo with the new url+commit, is that accurate?

jdbranham:
I guess rather than trying to keep compatibility with 4.x, I can just start the new repo with 5.x compatibility. Savvy users may still be able to use the updated plugin with 4.x installations.

PR #323, opened 2018-10-24
Added Diagram v1.4.5

jdbranham:
Let's try this again!
I fixed build script so we won't forget files in the dist folder anymore.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,1 @@
 node_modules
-/.project
--- a/repo.json
+++ b/repo.json
@@ -1172,7 +1172,7 @@
         },
         {
           "version": "1.4.5",
-          "commit": "80792ad1c95f8cfa48a6179427e2e57918ca7040",
+          "commit": "7119fb361566d8a37deb4ecca64896a5c36d17b6",
           "url": "https://github.com/jdbranham/grafana-diagram"
         }
       ]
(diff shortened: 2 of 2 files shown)


torkelo:
The panel sets the height of the outer panel element in a way that it should not:
https://github.com/jdbranham/grafana-diagram/blob/master/src/diagramControl.js#L631 

ctrl.height is max height available for inner elements. You should not touch the outer panel element as this mess with the panel container. 

image

Resizing a panel does also not any effect on the diagram dimensions.

torkelo:
seems to be related to this https://github.com/jdbranham/grafana-diagram/issues/87

midnightconman:
@torkelo Can you give us more information on the issue? I have verified this plugin with grafana 5.4.2

torkelo:
It does not respect its height and set it’s height in a wrong way. The v5 support issue is also still open

jdbranham:
thanks @torkelo I updated the css line you referenced, and closed a few issues. Grafana v5 compatibility is also ready.

midnightconman:
@torkelo Can you take a look at this? Some people are eager to see this upgrade available via grafana plugin distribution.

torkelo:
It's a bit better maybe, but still does not adapt the panel height, like most panels. 

Is there a reason it's fixed height? you can use ctrl.height (on the panel controller) to get the available panel height. Use use flex box css layout to autmatically size the diagram and legend components. 

But it's ok to publish, not at least it does not change grafana panel container dimensions.

alexanderzobnin:
@jdbranham grafana.com/plugins now supports change log, so you can add all notable changes into `CHANGELOG.md` in the root of repo in order to tell other people what's new.

midnightconman:
Thank you :smile:

Maybe in 1.4.6 we can look at the enhancements you listed.

PR #325, opened 2018-10-24
added kumaravel29's sankey-panel

petemyron:


briangann:
review is in progress, there are some changes required to make this plugin work "out of the box"

briangann:
* You'll need to update the plugin id in your src/plugin.json to match:
`"id": "kumaravel29-sankey-panel",`

* The documentation shows version 2.0.0, however the src/plugin.json says it is 1.1

* there's a typescript file `src/plugin_loader.ts` that appears to not be used

* in src/module.html, there is a spelling error `id="sankey_daigram"`, doesn't appear to be used anywhere though

I'll fork the repo and submit a PR with changes that will let it run without modifying grafana itself.

briangann:
@petemyron which datasource does this work with? I'm looking for some sample data that will work with it. A datasource + query example would be great.

I've created mock data that renders but would like to use with a real datasource.

marcusolsson:
Closing due to inactivity.

PR #327, opened 2018-10-31
Instana datasource plugin 1.2.2

schmidtl:
* fix metric querying for past time windows
* fix metric extrapolation for Singlestat and Table displaying


torkelo:
published!

PR #337, opened 2018-11-15
Instana datasource plugin 1.3.0

schmidtl:
- custom legend
- bugfix for failing focus queries on newly created dashboards


daniellee:
Merged and published

PR #343, opened 2018-12-13
Update flant-statusmap-panel version to 0.0.4

diafour:
A hotfix for a bug introduced in 0.0.3 version.

- Fix display of multivalues buckets as an empty cell


daniellee:
Merged and published.

PR #346, opened 2018-12-16
Update larona-epict-panel to 1.0.1

LucasArona:
Updated plugin larona-epict-panel to 1.0.1


daniellee:
Merged and published.

PR #347, opened 2018-12-17
Update parity report panel to v1.2.0

zuburqan:
Update version to v1.2.0

Changes include:

- Ability to configure alias key under options that gets extracted from the JSON response
- Updated readme
- Updated options screenshot


daniellee:
The version in the [plugin.json](https://github.com/zuburqan/grafana-parity-report/blob/master/dist/plugin.json#L7) is incorrect - 1.0.1 instead of 1.20.

zuburqan:
Fixed. I have submitted another PR

Thanks
Z

daniellee:
Fixed by #354

PR #351, opened 2018-12-19
update discrete to 0.0.9

ryantxu:
this fixes two key bugs:
https://github.com/NatelEnergy/grafana-discrete-panel/issues/62
and
https://github.com/NatelEnergy/grafana-discrete-panel/issues/74

It also has lots of build improvements like: using webpack and releasing with release-it


daniellee:
Looks good! Merged and published.

PR #356, opened 2019-01-04
Blendstat panel

farski:


daniellee:
There is a small error that needs to be fixed before I can start reviewing. The plugin id in the [plugin json file](https://github.com/farski/blendstat-grafana/blob/master/dist/plugin.json#L4) is incorrect. It should be the same as one in the PR: `farski-blendstat-panel`.

farski:
@daniellee Should I be fixing this merge conflict?

farski:
@daniellee Sorry to bother about this, any idea when it might be available in the marketplace?

daniellee:
@farski really sorry for the delay - this looks fine except for one thing:

the paths to the editor tabs in the module.ts file are incorrect. It should not be `public/plugins/blendstat/...` as the plugin name is now farski-blendstat-panel:

```typescript
this.addEditorTab('Options', 'public/plugins/farski-blendstat-panel/editor.html', 2);
this.addEditorTab('Value Mappings', 'public/plugins/farski-blendstat-panel/mappings.html', 3);
this.addEditorTab('Blending', 'public/plugins/farski-blendstat-panel/blending.html', 4);
```

farski:
@daniellee I think that's squared away, thanks!

PR #368, opened 2019-01-29
update snuids/grafana-radar-panel v1.4.2

snuids:
v1.4.2 (30/Sep/2018)
Alias option added in order to rename the serie for SQL datasources


briangann:
looks good, verified working

PR #375, opened 2019-02-07
Update clickhouse-grafana to 1.8.1

hagen1778:
Please, see release updates https://github.com/Vertamedia/clickhouse-grafana/releases/tag/1.8.1

New features:
* Add `timeFilterByColumn` macro 

Fixes:
* add requestId to queries so that abandoned one are cancelled 
* bug with parentheses in `$unescape` macros
* bug with treating string as numbers in table view

cc @nvartolomei


daniellee:
I saw something in the first PR about being able to use the new macro in query variables but it doesn't work there (think it is just me not understanding). But otherwise everything looks fine and the macro works fine in a normal query.

Merged and published

PR #382, opened 2019-02-17
thruk-datasource: update to 1.0.3

sni:
1.0.3  2019-02-15
    - support aggregation functions
    - convert hash responses into tables
    - support timeseries based panels

### DIFF
--- a/repo.json
+++ b/repo.json
@@ -1963,6 +1963,11 @@
           "version": "1.0.2",
           "commit": "3232fa771ae5c38e55695243a056540a76bd0bc3",
           "url": "https://github.com/sni/grafana-thruk-datasource"
+        },
+        {
+          "version": "1.0.3",
+          "commit": "a3977184331020124b486309e320a08ab2a1f186",
+          "url": "https://github.com/sni/grafana-thruk-datasource"
         }
       ]
     },


daniellee:
Some big new features in this version. I had some trouble getting the example query working for timeseries:

`SELECT count() state, state FROM /hosts`

but this worked but had no legend keys:

`SELECT count(state) FROM /hosts`

Am I doing something wrong? I tried switching the order of the fields around but couldn't get it to work:

image

sni:
`SELECT count() state, state FROM /hosts` is correct and should return the number of hosts grouped by state. I know it looks a bit funky. Internally it gets converted to `SELECT count(state), state FROM /hosts` but i couldn't get the query editor to work like that.

From the screenshot i assume you are trying with beta6, let me try to reproduce that. It should work with grafana 5.

sni:
ok, i just checked with the grafana/grafana:6.0.0-beta2 docker image and it also works. The reason why you don't see labels is that `count()` and `host_name` must be two separate columns.
It works if you choose it like this:

screenshot 2019-02-22 at 17 03 59

sni:
but i agree, the query editor could be made easier in that area.

daniellee:
Thanks for the explanation. Merged and published.

PR #395, opened 2019-03-09
update grafana-radar-panel: Few graphical options added

snuids:
Merged @fremag modifications: New chart options: min, max, step, animation

### DIFF
--- a/repo.json
+++ b/repo.json
@@ -2485,6 +2485,13 @@
           "commit": "bbc9761b0ea59f9ed50363dacaa2ece6b8b293b2",
           "url": "https://github.com/snuids/grafana-radar-panel"
         }
+        ,
+        {
+          "version": "1.4.4",
+          "commit": "34713c10ac816b9752c79e43a4b3f630d58e9953",
+          "url": "https://github.com/snuids/grafana-radar-panel"
+        }
+        
       ]
     },
     {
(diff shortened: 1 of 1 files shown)


alexanderzobnin:
@snuids 
1. There's a version mismatch between repo.json and plugin.json: `Version in repo.json (1.4.3) doesn't match plugin.json (1.4.2)`
2. You should add a new version into the 'repo.json` rather than change an old one.

snuids:
@alexanderzobnin

I restored the previous version.

спасибо,

alexanderzobnin:
@snuids version in `plugin.json` is still outdated: `Version in repo.json (1.4.3) doesn't match plugin.json (1.4.2)`

snuids:
@alexanderzobnin 

I am probably missing something here. I guess that the repo.json is OK. The plugin.json in my repo is also 1.4.3. Is there something more that I should do ?

Cheers,

briangann:
@alexanderzobnin looks like repo.json and plugin.json match, should be fine to publish.

alexanderzobnin:
@snuids check `plugin.json` in the dist/ folder. It seems, you forgot to build it:
https://github.com/snuids/grafana-radar-panel/blob/7ee4013b618875710b2c767542095ce9ab91a645/dist/plugin.json#L29

snuids:
@ alexanderzobnin

Hi Alexander. You were right since the beginning. I lazily tried to update the plugin in the repo without building it.

Thx,

Cheers

snuids:
Re opened as a new request

PR #396, opened 2019-03-11
Update flant-statusmap-panel version to 0.1.0

diafour:
- Tested with Grafana 6.0.0
- Tested with InfluxDB and Mysql datasources
- Add initial support for display annotations
- Add example for k8s statuses (thanks, @vrutkovs)
- Fix hanging on big values
- Fix horizontal spacing = 0
- Fix for "Object doesn't support property or method 'remove'"
- Fix card width for targets with different datapoints count


alexanderzobnin:
Looks good to me, merging. If you want to provide better change log experience, put it into `CHANGELOG.md` in the root of repo. Thus, it will be published and available under `Change log` tab at the plugin page.

PR #398, opened 2019-03-12
Pubish flowcharting 0.2

algenty:
Fisrt publish of my plugin.
Based on mxgraph libs. I know i need to clean code for next versions
its a beta need beta tester

Best reguards,
Arnaud

Sorry for my powerless english


algenty:
Retry

PR #414, opened 2019-04-01
Instana datasource plugin 2.2.1

schmidtl:
* make proxy mode configurable
* increase fetch size for websites and applications

### DIFF
--- a/repo.json
+++ b/repo.json
@@ -2474,6 +2474,11 @@
            "version": "2.2.0",
            "commit": "7babdbc3c0123bfbe74950ea4abfffa63739433d",
            "url": "https://github.com/instana/instana-grafana-datasource"
+        },
+        {
+           "version": "2.2.1",
+           "commit": "2cf6e937b530fa4c49023844ff9f7692067dfe0e",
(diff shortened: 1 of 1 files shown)


daniellee:
Merged and published.

Some feedback - if you are going to use Grafana's proxy mode then you can also encrypt the API key if you want to add some extra security. Instead of using the jsonData blob, you can use the secureJsonData blob that encrypts the data. This blob is only decrypted on the backend and is never sent to the frontend again.

Docs here: http://docs.grafana.org/plugins/developing/auth-for-datasources

It requires more work on the frontend as you have to allow the user to reset the api key. Here is an example for the Postgres password field:

image

In the config partial, you can use `ctrl.current.secureJsonFields.yourField` to see if it has a value or not - returns true if the field has a value.

Example of a reset button from the Azure Monitor datasource that hides the input field when the secureJsonData field has a value:

https://github.com/grafana/grafana/blob/master/public/app/plugins/datasource/grafana-azure-monitor-datasource/partials/config.html#L50-L68

schmidtl:
Thanks @daniellee I'm going to check this and figure out how to attain that while still supporting our old configurations

PR #418, opened 2019-04-20
Radar Panel v1.4.4 (20/Apr/2019)

snuids:
# Radar Panel
## v1.4.4 (20/Apr/2019)
- Autoscale option added
- Tick background can be hidden


alexanderzobnin:
@snuids Looks good, but I see that version in `plugin.json` is `1.4.3`. Is it right?

snuids:
I made a new version 1.4.4 since my last pull request attempt.

So there should be two tags with two commits that match the two versions.

Cheers,

alexanderzobnin:
Aha, OK, I see.

snuids:
Good thx :-)

PR #427, opened 2019-05-10
Instana datasource plugin 2.2.3

schmidtl:
- add groupbyTagSecondLevelKey to applications call.http.header


daniellee:
Merged and published

PR #428, opened 2019-05-10
Update BT alarm-box and status-dot plugins

jonasbleyl:
* Fixes "Values" tab not showing up in Grafana 6


daniellee:
Status dot and alarm box are published now (and can confirm that the values section works again).

PR #429, opened 2019-05-14
Instana datasource plugin 2.2.4

schmidtl:
* fix custom legend labeling for infrastructure
* fix second level group reset
* dependency updates


daniellee:
Merged and published

PR #435, opened 2019-05-24
Update Multistat version to 1.2.3 

michaeldmoore:
Minor fix to correct init problem with older versions of Grafana


daniellee:
Merged and published

PR #439, opened 2019-06-04
BQ Data Source 0.6.0

avivl:
@daniellee


daniellee:
@avivl Thanks - started reviewing and will create some issues around some small bugs that I found.

avivl:
@daniellee I have updated it with a new version. any eta?

daniellee:
Merged and published. Found some issues but they are nothing new I think. I need to spend some time on digging a bit deeper to create good issues for them.

PR #443, opened 2019-06-24
Instana datasource plugin 2.3.1

schmidtl:
* increase search limit for offline snapshots to api default


daniellee:
Merged and published

PR #467, opened 2019-08-16
Add image renderer plugin

aocenas:
Add grafana-image-renderer plugin.
Is blocked by https://github.com/raintank/grafana-net/pull/913 right now.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,3 @@
 node_modules
 .vscode
+.idea
--- a/repo.json
+++ b/repo.json
@@ -3190,7 +3190,7 @@
       "versions": [
         {
           "version": "1.0.0",
-          "commit": "cabf35236854944f7105af8cc422d4f275af4e40",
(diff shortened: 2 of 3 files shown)


aocenas:
@marefr based on what @briangann told me there is a pending flag we can use when registering it. Should make it not visible in the UI I assume and I hope it will make it still downloadable by the id from CLI but need to check that.

> I guess we could have tagged it with alpha/beta

Is this something existing that grafana.com understands? Or do you mean just in the semver string?

aocenas:
Looks like pending should work, should not be listed but should still be installable by the CLI. So something like:
```
POST
https://grafana.com/api/plugins
{
  "slug": "grafana-image-renderer",
  "pending": true
}
```
Should work.

marefr:
> Is this something existing that grafana.com understands? Or do you mean just in the semver string?

I was referring to the semver string. We can skip this for now since we're publishing it in pending mode.

PR #470, opened 2019-08-26
upgrade `goshposh-metaqueries-datasource` to `0.0.3`

Gauravshah:
added support for grafana 6.3 as `ds.name` is removed in commit : https://github.com/grafana/grafana/commit/1d7bb2a#diff-894468679ebfe67fc2c22f7575f9e2c2

### DIFF
--- a/repo.json
+++ b/repo.json
@@ -2730,6 +2730,11 @@
       "url": "https://github.com/GoshPosh/grafana-meta-queries",
       "versions": [
         {
+          "version": "0.0.3",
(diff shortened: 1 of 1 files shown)


daniellee:
Hi @Gauravshah this plugin still has a lot of problems but publishing as it is an improvement over the existing version (which usually crashed the browser for me).

- Adding a query C and then hiding it doesn't work:
    image

- Timeshifting queries are always canceled:

    image

Lots of JavaScript errors that pop up regularly: 
image

image

Gauravshah:
thanks @DanCech created an issue to folllowup

PR #474, opened 2019-09-06
Update akumuli-datasource plugin to 1.3.9

Lazin:
Add annotations support.

### DIFF
--- a/repo.json
+++ b/repo.json
@@ -2718,8 +2718,8 @@
       "url": "https://github.com/akumuli/akumuli-datasource",
       "versions": [
         {
-          "version": "1.2.8",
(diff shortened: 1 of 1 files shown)


daniellee:
@Lazin The link to the image for annotations in the readme is a relative link:

`Dashboard alert`

Can you change it to be an absolute link instead otherwise it it will be broken on grafana.com.

Have you got an example of an annotations query on the test server (http://206.189.27.155:8181/)?

Lazin:
@daniellee Fixed the link. Also added cron-job that creates event every 30-minutes on test server.
You need to specify 'alarm' as an event name to see them.

PR #490, opened 2019-10-01
Instana datasource plugin 2.4.1

schmidtl:
* add application metric categories
* extend snapshot caching

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,3 @@
 node_modules
 .vscode
+.idea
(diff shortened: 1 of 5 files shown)


schmidtl:
Hi, what is the status for the update?
Thank you.
/Lars

schmidtl:
Hi, what is the current status for our plugin update?
Thank you.
/Lars

briangann:
reviewing now

PR #491, opened 2019-10-02
Multistat update to 1.2.4 - fix for grafana 6.4.x table data response error.

michaeldmoore:
Grafana 6.4.0 update introduced a bug which removed the type='table' value for non-timeseries data queries.  Multistat, which only works with table-formatted data relied on this property.  This minor version release adds an alternative method of determining the query type (namely the existence of a columns property which only table-formatted queries produce.


daniellee:
Looks good to me

PR #492, opened 2019-10-03
OpenNMS Helm 4.0.0

RangerRick:
Updated OpenNMS Helm to `4.0.0`.

If you wish to try out the datasources, you can use `https://demo.opennms.org/opennms/` as your URL root with basic auth, username:pass `demo`:`demo`.

This is a pretty big update with a bunch of refactoring and a ton of new features.  Release notes reproduced here:

## General

Helm 4 now requires Grafana 6.0 or higher.

## Entities Data Source

The Fault Management Data Source is now the Entities Data Source.

In addition to querying alarms, it can now query nodes as well.
This is useful for filtering in variables, or can be used for just viewing and filtering nodes from within Grafana.

For details on upgrading, see [the Helm documentation](http://docs.opennms.org/helm/releases/latest/helm/latest/installation/upgrading.html).

## Flow Data Source

- Enhancements have been made to support Horizon 25's addition of "top N" queries for conversations, hosts, and applicattions.

## Filter Panel

A new panel has been added that allows you to configure a series of variables to put in a dashboard.
This panel coordinates with any panels in the dashboard using the Entities Data Source to automatically apply filters matching those variables.

In the future this will be enhanced to do more complicated set operations than are possible using the built-in datasource variables.

## Alarm Table

- Alarm table code has been sync'd with enhancements from the upstream Grafana table panel.
- The new default columns and formatting for the alarm table have been updated to be more useful.
- It is now possible to configure the alarm table to not automatically refresh when new data is updated if you have navigated away from the first page of the list. When new data arrives in the background, a refresh icon will appear in the page bar.
- String-based columns (like log message) now get a mouseover with the full text of the column.
- Severity columns have been enhanced, including basic support for themes.
- A new column type (`checkbox`) has been added that will render a boolean value as either a checkmark or empty string.
- Support has been added to tag a specific alarm as the "root cause" in a situation when providing feedback in the Alarm Details screen.
- A number of UI cleanups have been made to the Alarm Details screen, fixing word wrapping, resizing when the browser is resized, and more.
- The alarm table can now export CSV and Excel files.  Go to `<panel title>` -> `More...` -> `Export CSV` or `Export Excel`.

### DIFF
--- a/repo.json
+++ b/repo.json
@@ -685,6 +685,11 @@
       "url": "https://github.com/OpenNMS/opennms-helm",
       "versions": [
         {
+          "version": "4.0.0",
+          "commit": "12cc57953ac69232d992d4476856ef28f9266bc9",
+          "url": "https://github.com/OpenNMS/opennms-helm"
+        },
+        {
           "version": "3.0.1",
           "commit": "94fc9a3de676b4d993114cff119dec425fa502af",
           "url": "https://github.com/OpenNMS/opennms-helm"


RangerRick:
Looks like 6.4 broke this, so I'm gonna close it for now and reopen when I've figured out why.

PR #498, opened 2019-10-17
Instana datasource plugin 2.4.2

schmidtl:
* extend second level grouping
* fix version check
* fix proxy enabling on datasource configuration


daniellee:
Merged and published

PR #502, opened 2019-10-28
Instana datasource plugin 2.4.3

schmidtl:
* fix rollup and granularity calculation
* fix rollup and granularity selection


daniellee:
Merged and published.

Seeing some behavior around roll-ups which seems a bit strange.

1. Starting with a 5m rollup:
    image
2. Changing to 10m:
image
3. Changing back to 5m:
image
4. If I save and refresh the page, then it looks like step 1 again.

Is this a bug or a feature?

schmidtl:
Hi Daniel, 
thanks for the fast merge and the feedback. I will check the rollup behavior.
Best regards
/Lars

PR #513, opened 2019-11-12
Instana datasource plugin 2.4.4

schmidtl:
* fix selection for "-- No Application Filter --"
* fix rollup & granularity selection
* fix timeshift validation


schmidtl:
👍

PR #519, opened 2019-11-16
Updated larona-epict-panel to 1.2.2

LucasArona:
- New feature: A tooltip is now available on the orbs with the current value.
- New feature: You can specify a rotation angle for the boxes.
- New feature: You can use variables in the Metric Name field of the boxes, so you can easily change the metrics used by your boxes.


daniellee:
Merged and published

PR #521, opened 2019-11-18
Update corpglory-progresslist-panel to 1.0.5

corpglory-dev:
#### Added
- "Opacity" option for bars
- Tooltip with current item values

#### Changed
- Title and progress bar in one line (optional)


briangann:
merged and published!

PR #529, opened 2019-12-02
Updated Alexandra-trackmap-panel to version 1.2.3

Drazair:
Fixed a problem when multiple plugins used leaflet. 
Added more query examples to the readme.


briangann:
merged and published

PR #534, opened 2019-12-18
added opfa-category-heatmap-panel plugin

Petro36:
added opfa-category-heatmap-panel plugin

### DIFF
--- a/repo.json
+++ b/repo.json
@@ -3547,6 +3547,18 @@
       ]
     },
     {
+      "id": "optionfactory-categoryheatmap-panel",
+      "type": "panel",
+      "url": "https://github.com/optionfactory/category-heatmap",
+      "versions": [
+        {
+            "version": "1.0.3",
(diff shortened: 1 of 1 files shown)


briangann:
Can you update it from using:

"id": "optionfactory-category-heatmap-panel"

to this (without the dash between category and heatmap)

"id": "optionfactory-categoryheatmap-panel"

The id/name is long, but will pass through our naming verification.

briangann:
Looks good so far. You'll need to provide a "dist" with the fully built plugin either in master, or in a branch, and use that as the commit hash.

Since you're all setup with @grafana/toolkit and also circleci, you could use the approve/publish method to create releases.

the pie-chart plugin has an example for circleci: https://github.com/grafana/piechart-panel/blob/master/.circleci/config.yml

marcusolsson:
Closing due to inactivity.

PR #535, opened 2019-12-19
Update akumuli-datasource to 1.3.10

Lazin:
This version adds ability to set time shift per series (grafana can only set time shift for the entire panel, but setting it per series can be useful to compare different time intervals, e.g. this month with the previous one).

Simple test:
- set up datasource using url: http://206.189.27.155:8181
- create panel with one query: parameters are `Metric` set to 'http.total', `Tags` set to 'endpoint=Gap.com'
- duplicate the query and edit it by setting `Time Shift` to '10m', you should see that the second graph is shifted back in time.


briangann:
merged and published!

PR #558, opened 2020-01-20
Instana datasource plugin 2.5.1

schmidtl:
* fixed group by application.name behavior
* fixed max windowSize calculation
* adjust min rollup calculation similar to ui


daniellee:
LGTM

PR #560, opened 2020-01-29
Updated Alexandra-Trackmap-Panel to version 1.2.4

Drazair:
Added option for automatically center of the map based on data points.
Added warnings if data is invalid.


Drazair:
Hey, I needed the old versions back, expect the ones before 1.2.2 since they contain a security vulnerability.

daniellee:
Published version 1.2.2 and 1.2.4.

PR #562, opened 2020-02-01
Update scadavis-synoptic-panel to version 1.0.3.

riclolsen:
Added "Auto Resize" config option.
Supported Grafana Templating Variables.
More efficient data update and display rendering.
Generic tagging (@1, @2, @3, ...) compatible with the Power BI plugin.
Better responsiveness to configuration changes.


riclolsen:
Ok, I will resubmit as you recommended. Thanks, Alexander.

PR #572, opened 2020-02-28
HDS Datasource plugin v1.0.1

dschowta:
Hi, 
This is update to the existing open source [HDS datasource plugin](https://github.com/linksmart/grafana-hds-datasource)  with following minor changes

- Support latest version of [HDS](https://github.com/linksmart/historical-datastore)
- Link compatibility matrix
- Bug fixes

This plugin can be tested with a fake HDS server (with continuously growing dummy senML data) by running following command:
```
docker run -p 8085:8085  linksmart/hds -demo -conf /conf/docker.json
```

Thank you.

### DIFF
--- a/repo.json
+++ b/repo.json
@@ -2963,8 +2963,8 @@
       "url": "https://github.com/linksmart/grafana-hds-datasource",
       "versions": [
         {
-          "version": "1.0.0",
-          "commit": "9cb2c740e78d1b09b7e920271689281fd52916f7",
+          "version": "1.0.1",
+          "commit": "36388a6bc4a94d197bf90513248e9df564323a7f",
           "url": "https://github.com/linksmart/grafana-hds-datasource"
         }
       ]


daniellee:
We made some large changes in 6.7, removing Angular services and replacing them with React ones . We changed the backendSrv to be fetch based and this is breaking your code somehow. I'll get back to you with a suggested fix (or maybe it is something that has to be fixed in Grafana).

daniellee:
I did some more testing and found that the issue is nothing to do with the 6.7 release.

- [This GET request](https://github.com/linksmart/grafana-hds-datasource/blob/master/src/datasource.js#L131) has a `data` parameter. This is incorrect - HTTP GET requests cannot have a body. They have query parameters. I'm not sure how this ever worked. It should either be changed to a POST request or the `data` options should be changed to `params`. `params` will then automatically be converted into a querystring and added to the url.
- I'm also seeing an infinite loop when querying with the previous version of this plugin but I think that is fixed now in https://github.com/linksmart/grafana-hds-datasource/commit/f4e1a369d41a7114c61b32e9495c5cb4de6f65ba?

dschowta:
Thank you for the comments. The commit https://github.com/linksmart/grafana-hds-datasource/commit/0b17f592727d0b2fdbaa99c1429f6f7a63100ff1 removes body in HTTP GET requests.

dschowta:
you are right about the commit
>I'm also seeing an infinite loop when querying with the previous version of this plugin but I think that is fixed now in linksmart/grafana-hds-datasource@f4e1a36?

daniellee:
Thanks for the fixes - works great now again.

PR #580, opened 2020-03-15
Add updated scadavis-synoptic-panel to version 1.0.4.

riclolsen:
Added "Load SVG from local file" config option.
Better auto-resize handling.


riclolsen:
I have found some problems here.

PR #582, opened 2020-03-21
Add SCADAvis Panel update 1.0.4.

riclolsen:
Added "Load SVG from local file" config option.
Better auto-resize handling.
Compatible also with Grafana 6.7.


daniellee:
Merged and published!

PR #584, opened 2020-03-23
Add kdb+ datasource v1.0.0

DGaquaq:
This is a new data-source plugin to enable connection to kdb+ historic and real-time databases over WebSockets. It has the ability to build custom queries over the 'Built Query' panel, or run free-form queries/functions, both supporting conflation and grouping. Tabular visualisations are also supported in both panel types.

Test:
 - Download [latest version of kdb+](https://kx.com/connect-with-us/download/) and setup as [described by kx](https://code.kx.com/v2/learn/install/).
 - Open a listening port and set WebSocket message handler as shown in README.md:

```
\p 6990
.z.ws:{ds:-9!x;neg[.z.w] -8! `o`ID!(@[value;ds[`i];{`$"'",x}];ds[`ID])}
```

 - Connect grafana data-source as described in README.md (Host: `localhost:6990`).
 - Save & Test
 - Examples of test data sets are [available from the kx github](https://github.com/KxSystems/kdb)

```
/ 10000 sample market trades
q)nt:10000
q)t:`time xasc ([]sym:nt#`FDP;time:09:30t+nt?06:30t;price:100+(nt?100)%100;size:100*1+nt?10)
/ time is required in timestamp format
q)update time:.z.d+time from `t
/ in-memory table should have both grouped and sorted attributes
q)update `s#time,`g#sym from `t
```

 - Navigate to new Dashboard -> add Query
 - Under 'From' select 't'
 - Under 'Time Column' select 'time'
 - Under 'Select' select 'price'
 - Set the time-range in Grafana to last 24 hr

### DIFF
--- a/repo.json
+++ b/repo.json
@@ -3766,13 +3766,13 @@
       ]
     },
     {
-      "id": "aquaq-kdb-websocket-adaptor",
+      "id": "aquaqanalytics-kdbadaptor-datasource",
       "type": "datasource",
       "url": "https://github.com/AquaQAnalytics/grafana-kdb-datasource-ws",
       "versions": [
         {
           "version": "1.0.0",
-          "commit": "e5cb3dbaf922a0e2f5832f4310a66c9e30eeb24c",
+          "commit": "f1c15ddc0eb9761f622cf00ccfc6dd478bf5561c",
           "url": "https://github.com/AquaQAnalytics/grafana-kdb-datasource-ws"
         }
       ]
--- a/repo.json
+++ b/repo.json
@@ -3771,8 +3771,8 @@
       "url": "https://github.com/AquaQAnalytics/grafana-kdb-datasource-ws",
       "versions": [
         {
-          "version": "1.0.0",
-          "commit": "f1c15ddc0eb9761f622cf00ccfc6dd478bf5561c",
+          "version": "1.0.1",
+          "commit": "9b39310d7254d5bb4f2d28f9ca48320ac5df1140",
           "url": "https://github.com/AquaQAnalytics/grafana-kdb-datasource-ws"
         }
       ]
--- a/repo.json
+++ b/repo.json
@@ -3772,7 +3772,7 @@
       "versions": [
         {
           "version": "1.0.0",
-          "commit": "f1495b10ecc73e0f5a99707b18ac13f1167824f5",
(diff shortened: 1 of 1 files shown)


CallumOrmerod:
Just adding a quick note to this to say that we're (AquaQ) are planning to actively develop this datasource and are currently working on a version two which will include features like:

- Support for Grafana Variables
- Support for multiple grouping fields from within the query builder
- Support for TLS 1.1 (kdb+ version 4.0 - just released)
- Wrapper for .z.ws to allow multiple websocket processes to connect to the same kdb+ instance.

daniellee:
@DGaquaq thanks for the instructions - they were very helpful. I got stuck on the last step though and it's probably me doing something stupid:

image

I ran the following command via q on the commandline and it seems to return data:

image

DGaquaq:
> @DGaquaq thanks for the instructions - they were very helpful. I got stuck on the last step though and it's probably me doing something stupid:

By time range in grafana I meant the time range one selects at the top right of the screen. The function I gave will generate data between 09:30 and 16:00 of the day it is run, so if you're viewing with the default time range (last 15 min I believe) and viewing later it won't be visible. I have attached image that should help.

grafana time range

Let me know if you're still having issues, thanks!

daniellee:
@DGaquaq I tested again from the start and I must have messed something up the first time. Now I can see data if I change the time range to look one day into the future - guess that function generates data from 9:30 today. I'm seeing some strange behavior though when I zoom in - points are missing. I would understand this if I had zoomed into look at a 1m interval but I'm looking at 2 hours in the example below and most of the points are missing.

Here is looking at the whole day:

image

When I zoom in:

image

DGaquaq:
This appears to be a timezone issue (we have been developing and testing in UTC time up to now). We are looking at it now, will have a fix out by the end of the day. Thanks for the help!

DGaquaq:
Hi Daniel,
We have fixed the time localisation bug; kdb+ times will always be interpreted as UTC so if the dashboard is not set to UTC time the data will be offset by the user's timezone (for yourself 09:30 will become 11:30; grafana-side request and rendering of this data has been corrected).
For the next version we aim to add column-specific time offsets for users who capture/store data in non-UTC time zones.
Thanks

daniellee:
Merged and published: https://grafana.com/grafana/plugins/aquaqanalytics-kdbadaptor-datasource

PR #585, opened 2020-03-24
Update Multistat to v1.2.7 (minor bug fix)

michaeldmoore:
NON_URGENT UPDATE.

This release fixes an odd problem in which multiple groups of records had identical labels.  The aggregation function failed to consider the group  value when identifying duplicate records.
All fixed.


daniellee:
Merged and published.

PR #597, opened 2020-04-12
Update Flowcharting 0.7.1

algenty:
Hi teams,

Only some fix at this time, 
Thx a lot

### Fixed  
  - shape stills select when deleting map.
  - select list (typeahead) in fields 'whats' and map values for string type in rules
  - Level State with range or value mapping.
  - Lock always true in flowchart option [(issue #68)](https://github.com/algenty/grafana-flowcharting/issues/138) 
  - Fix Remove link
  - Fix range value Min and Max range when 'from' or 'to' is empty
  - Fix 'With' field in text mapping, empty after clone or save/reload

### DIFF
--- a/repo.json
+++ b/repo.json
@@ -3372,6 +3372,11 @@
           "version": "0.7.0",
           "commit": "adbfe9849a3ad2c85551fe347f2215b59340a0a1",
           "url": "https://github.com/algenty/grafana-flowcharting"
+        },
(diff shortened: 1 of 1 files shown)


daniellee:
Published and updated on the play site

algenty:
You are the best,
Thanks a lot.

PR #598, opened 2020-04-12
Update Annunciator to v1.0.5, adding new 'Shock & Awe' alarm indicator

michaeldmoore:
Added option to display unmistakable 'Shock and Awe' display using CSS animation to display threshold violations

### DIFF
--- a/repo.json
+++ b/repo.json
@@ -2393,7 +2393,7 @@
         },
         {
           "version": "1.0.5",
-          "commit": "3706498e0eabaf46b4e501eade3b445c0ef24a95",
+          "commit": "3796550c2d733b777f85f1536b086b1ade33b75b",
           "url": "https://github.com/michaeldmoore/michaeldmoore-annunciator-panel"
         }
       ]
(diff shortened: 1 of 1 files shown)


daniellee:
@michaeldmoore you missed bumping the version here: https://github.com/michaeldmoore/michaeldmoore-annunciator-panel/blob/master/dist/plugin.json#L13

michaeldmoore:
Wow - good catch.  Very impressive catch actually.

Recompiled and re-committed with new hash (
3796550c2d733b777f85f1536b086b1ade33b75b).   updated the repo.json patch
accordingly.

Thanks Daniel
Michael Moore

daniellee:
Sorry for the delay - merged and published

michaeldmoore:
No problem, thank you for getting to this.   Just checked out Version 7
(beta) - looks really good too, though I shall have to reorganize the
settings forms to make them less wide.

Cheers
M

daniellee:
Will talk to the frontend team to see if they can make some tweaks. Your two plugins look mostly ok but other plugins are affected more by the UI changes.

michaeldmoore:
I'm getting a report that with g7, my multistat panel isn't recognizing a
table-formatted data set as being a table.  I haven't been able to
reproduce it here, but have requested the person reporting this to send me
a copy of the JSON data from the query inspector.  I'm looking for the
type:"table" value - is this still the proper way to tell if a data set is
table-formatted as opposed to a time-series?

michaeldmoore:
The guy was using a SQL Server data source, something I can't test here
today.

PR #600, opened 2020-04-19
Update Multistat panel plugin to version 1.3.0

michaeldmoore:
Adding support for an array of recoloring rules, to intelligently override the colors of selected bars.


daniellee:
LGTM. I tested this on the latest master which has a lot of changes in the UI. This plugin's options look pretty good in the new panel edit on the side. Some plugins with lots of tabs and options will have problems in 7.0 but as you have used the right components, you shouldn't have to change anything:

image

PR #611, opened 2020-04-29
Add Matomo Tracker panel

thiagoarrais:
[Matomo](https://matomo.org/) is a web analytics platform. The Matomo Tracker panel is a Grafana Panel that allows Grafana Dashboards to be tracked by Matomo. It works much like a tracking pixel and is not intended to have any visible presence in the dashboard.

This plugin may help in https://github.com/grafana/grafana/issues/12234 and https://github.com/grafana/grafana/issues/3496

To setup a Matomo instance for testing, you can use the free Matomo hosted option. I've also achieved some success using [matomo-docker's nginx setup](https://github.com/matomo-org/docker/tree/master/.examples/nginx).

### DIFF
--- a/repo.json
+++ b/repo.json
@@ -4104,6 +4104,18 @@
       ]
     },
     {
+      "id": "thiagoarrais-matomotracking-panel",
+      "type": "panel",
+      "url": "https://github.com/thiagoarrais/grafana-matomo-tracking-panel",
+      "versions": [
+        {
+          "version": "0.2.2",
+          "commit": "1b334b8b0bd3b07909623dbdbdf560d969be0328",
+          "url": "https://github.com/thiagoarrais/grafana-matomo-tracking-panel"
+        }
+      ]
+    },
+    {
       "id": "grafana-strava-datasource",
       "type": "datasource",
       "url": "https://github.com/grafana/strava-datasource",


jordo1138:
I’ll test it out, this is a interesting approach as then as you programmatically build dashboards you could pick and choose quickly what gets tracks without tracking any other components of Grafana. It’s likely useful for teams who don’t want to track Grafana usage by feature which is what GA does now for Grafana but are interested in top used dashboards or for external facing dashboards

thiagoarrais:
Friendly ping.

@daniellee and @briangann: do we need anything else here? Is this pointing in the right direction? Just let me know!

thiagoarrais:
Just noticed that Grafana 7.0.0 has been released in the meantime. I've adapted the plugin to the 7.x series and rebased the PR. Please let me know if you need anything else.

briangann:
just checking - this plugin only works with a subscription to matomo correct?

thiagoarrais:
One needs to have access to some Matomo instance, yes. A self-hosted instance could be used, but there needs to be somewhere for the tracking data to go.

By the way, plugin id updated.

thiagoarrais:
Friendly ping here. Are there any blockers for this? Just let me know if you need anything.

marcusolsson:
Hey @thiagoarrais! We've been developing a tool to help us validate plugins. Would you mind entering the URL for this PR in the [plugin validator](https://grafana-plugins-web-vgmmyppaka-lz.a.run.app/), and fix the issues reported?

thiagoarrais:
@marcusolsson: the validator says the `skipDataQuery` property is not allowed. I'm using this to hide the query editor (just like the builtin Text Panel). Is this being deprecated? I don't see how can I achieve the same effect without it.

marcusolsson:
Hm, looks like you've discovered a gap in our plugin schema validation. I'll get this documented tomorrow!

thiagoarrais:
Fixed all the remaining issues and rebased.

thiagoarrais:
I see that now the validator gave us the green light. Just let me know if you folks need anything else.

marcusolsson:
[Published](https://grafana.com/grafana/plugins/thiagoarrais-matomotracking-panel)!

Thank you!

PR #618, opened 2020-05-11
Instana datasource plugin 2.7.2

schmidtl:
* fix json structure migration for Analyze application calls
* fix slo/sli naming


daniellee:
I tested with 7.0 beta3 as well. Merged and published.

PR #628, opened 2020-05-21
Add an image panel plugin

Dalvany:
A panel that can display an image by concatenating a prefix (the
beginning of an URL), a metric and a suffix.

### DIFF
--- a/grafana-plugin-repository.iml
+++ b/grafana-plugin-repository.iml
@@ -1,0 +1,9 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<module type="WEB_MODULE" version="4">
+  <component name="NewModuleRootManager" inherit-compiler-output="true">
+    <exclude-output />
+    <content url="file://$MODULE_DIR$" />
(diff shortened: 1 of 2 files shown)


marcusolsson:
Thank you @Dalvany for contributing your plugin to Grafana! I've tested your panel and have a few comments.

### Please fix

These are things that might negatively impact the user experience, and should be fixed.

**No required fixes**

### Please consider fixing

These are recommendations that will improve the quality of the plugin, but aren't required for publishing the plugin.

- I had to disable the tooltip to see the image. Otherwise I got: "Can't find time field for tooltip.”. I suggest disabling the tooltip by default. When enabling the tooltip, I suggest that “Include date” is disabled, by default. I think this will make it easier to get started.

Dalvany:
Hi @marcusolsson , thanks for the feedback. I made the requested changes and switch default options values for tooltip and date inclusion to false.

marcusolsson:
Thank you!

PR #630, opened 2020-05-24
Update Multistat to v1.4.0 (restyled tooltips plus clickable bar-links for drilldowns)

michaeldmoore:
Re-requesting change after (stupidly) deleting my forked patch of repo.json which cancelled the earlier pull-request.  Duh!

### DIFF
--- a/repo.json
+++ b/repo.json
@@ -3278,7 +3278,7 @@
         },
         {
           "version": "1.4.0",
-          "commit": "c7e5624359fd8cf95f2b93d840cfa8155d6bd993",
+          "commit": "0fae84cc0252dae635974d9e04d8e498bb313626",
           "url": "https://github.com/michaeldmoore/michaeldmoore-multistat-panel"
         }
       ]
--- a/repo.json
+++ b/repo.json
@@ -3278,7 +3278,7 @@
         },
         {
           "version": "1.4.0",
-          "commit": "b65fac7e2f1686f482cc802b7da7f89a22940c3e",
+          "commit": "13b3d1e4b627f07315952e2e1f75bcc987af76c0",
           "url": "https://github.com/michaeldmoore/michaeldmoore-multistat-panel"
         }
       ]
--- a/repo.json
+++ b/repo.json
@@ -3278,7 +3278,7 @@
         },
         {
           "version": "1.4.0",
-          "commit": "13b3d1e4b627f07315952e2e1f75bcc987af76c0",
+          "commit": "20b6f3a04e49b32c8d3ca66796426eff83738f1b",
           "url": "https://github.com/michaeldmoore/michaeldmoore-multistat-panel"
         }
       ]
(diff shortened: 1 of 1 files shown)


briangann:
seeing an error in the console output, haven't located exactly why it is being thrown:

```
angular.js:14800 Error: [$parse:syntax] Syntax Error: Token 'of' is an unexpected token at column 8 of the expression [Number of hours offset for UTC display] starting at [of hours offset for UTC display].
http://errors.angularjs.org/1.6.9/$parse/syntax?p0=of&p1=is%20an%20unexpected%20token&p2=8&p3=Number%20of%20hours%20offset%20for%20UTC%20display&p4=of%20hours%20offset%20for%20UTC%20display
    at angular.js:116
    at $r.throwError (angular.js:15358)
    at $r.ast (angular.js:15108)
    at Er.getAst (angular.js:16467)
    at Er.parse (angular.js:16450)
    at a (angular.js:16608)
    at Object.link (angular-strap.js:97)
    at angular.js:1383
    at It (angular.js:10619)
    at it (angular.js:10008) "<input type="number" class="gf-form-input width-3 ng-pristine ng-untouched ng-valid" data-placement="right" bs-tooltip="Number of hours offset for UTC display" ng-model="ctrl.panel.TZOffsetHours" ng-change="ctrl.render()" ng-model-onblur="">"
```

the editor has a couple of minor spacing/width problems:
image

image

otherwise looks and works great, if the utc bug is known i'll go ahead and merge/publish.

michaeldmoore:
Thanks Brian - this was my problem, so thanks again for spotting it.  Apparently, the bs-tooltip attribute MUST be surrounded by single quotes inside the double quotes.  One of the 37 such tags - one that I had just updated, obviously, was missing the inner single quotes. - duh!.

The new right-hand-side configuration layout has been a big change - I've had to squeeze everything horizontally, removing any unnecessary pixel to get things to fit in.  For me, at least, everything does fit in - at least, without word wrapping.  Maybe your machine has a larger system font definition.  Anyway, for this release, at least, I'd like to leave it all as-is, if you don't mind.

I'll have yet another, fairly significant version release in the coming couple of weeks - I'm adding support for multiple values as side-by-side bars and so far, it looks really good.  I'll also change the event registration using Panel.Events so the derication warning goes away too.  I will take another run through the configuration label widths too, to see if I can't make it even more beautiful.

One question, if you don't mind.  Is there a good way to tell if the dashboard is using the light or dark theme?  I have all the defaults set for the dark theme, which makes labels and things unreadable until they are changed with the light theme.  It would be nice to automate this if I can.

Thanks again,
Michael Moore

briangann:
one other thing i noticed, shouldn't this be `!==` instead of `!=`

```
<div class="gf-form" ng-if="ctrl.panel.DateTimeColName != ''">
```

for light/dark mode there are options to check that in the docs here: 
<https://github.com/grafana/grafana/blob/798d62c6057bb0faf8dd460785f5bfb32b5e6bc9/contribute/style-guides/themes.md>

PR #642, opened 2020-06-09
Update 'pr0ps-trackmap-panel' to v2.1.0

pR0Ps:
- Add option to zoom with scroll wheel
- Add option to set the default layer
- Add option to disable the layer switcher
- Fix small issue with zooming to fit data
- Fix snapshot rendering
- Update package deps


daniellee:
Merged and published

PR #644, opened 2020-06-10
Adding word cloud panel

raulsperoni:


marcusolsson:
Hey! Seems to be working as intended. Nice!

### Please fix

- The `name` property in the `plugin.json` determines the text of the panel in the UI. I'd suggest that you call it something more easier to read, maybe "Word cloud"?

- The README currently doesn't contain any instructions for end users. If you prefer to maintain two seperate READMEs, you could keep the dev-facing one in the root directory, and copy an end-user facing README to the `dist` directory.

### Consider fixing

- The defaults for the text and count fields aren't really useful. The user is likely not going to have fields called "Count" and "tweet_content.tags". One suggestion is to, by default, select the first field of type `string` for the text, and the first field of type `number` for the count. That way, chances are it'll work out of the box for more users.
  - Here's [an example](https://github.com/marcusolsson/grafana-treemap-panel/blob/50a42ee5b4f78b319bf71d3ce5defb73c912d650/src/TreemapPanel.tsx#L21-L29).
  - I wrote a [custom editor](https://github.com/marcusolsson/grafana-treemap-panel/blob/master/src/module.ts#L39-L49) for selecting fields, that might be interesting for you.

raulsperoni:
Thank you for your comments @marcusolsson ,

I think I've fixed the first two issues. Regarding the text ando count fields you're right. I've briefly tried your suggestion but I'm getting an error. Apparently "context" in FieldSelectEditor (I just took your code) is undefined:

`TypeError: Cannot read property 'data' of undefined`

Thank you.

PR #658, opened 2020-06-20
Add cloudspout-button-panel

derjust:
Please consider adding cloudspout-button-panel to the plugin repository.

It provides a simple panel that shows only one button - to integrate with any kind of HTTP/REST API:
* Support GET and POST HTTP verb
    * Adds no new javascript dependencies
    * Uses standard browser APIs and respects CORS
* Support API key via header X-API-Key or query parameter ?api-key
* Custom label text & Grafana template design

### DIFF
--- a/repo.json
+++ b/repo.json
@@ -4113,39 +4113,6 @@
           "url": "https://github.com/marcusolsson/grafana-hourly-heatmap-panel"
         }
       ]
-    },
-    {
-      "id": "marcusolsson-treemap-panel",
-      "type": "panel",
-      "url": "https://github.com/marcusolsson/grafana-treemap-panel",
-      "versions": [
-        {
-          "version": "0.1.0",
-          "commit": "f2ea6e9907a87bef6eb1a982d7ade3ff7cf16e75",
-          "url": "https://github.com/marcusolsson/grafana-treemap-panel"
-        }
-      ]
-    },
-    {
-      "id": "dalvany-image-panel",
-      "type": "panel",
-      "versions": [
-        {
-          "version": "2.0.0",
-          "commit": "342ccfaa188aef5112f067fad5317d0c9504a102",
-          "url": "https://github.com/Dalvany/dalvany-image-panel"
-        },
-        {
-          "version": "2.1.0",
-          "commit": "3063c83128315bd0b49b793fb9fa761751028f0c",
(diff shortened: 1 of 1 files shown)


marcusolsson:
Thank you @derjust for contributing your plugin to Grafana! I've tested your panel and have a few comments.

### Please fix

These are things that might negatively impact the user experience, and should be fixed.

- Image and GIF in README needs to be absolute URL rather than relative.
- Missing option names for Header/Query select, Color, and Method.
- The colors "Danger", "Inverse", and "Transparent" seems to be the same as the Primary?
- API Key hard-codes the query param and header that is used. I’d expect a button like this to support adding custom headers and query parameters. The user is asked to enter a (secret) API Key in clear text, which shouldn’t be encouraged. By supporting generic headers and params, the user can still add secrets, but it’s at least not actively encouraged.
 
### Please consider fixing

These are recommendations that will improve the quality of the plugin, but aren't required for publishing the plugin.

- Make the README heading the same as the title of the plugin, e.g. “Button panel”.
- `x-api-key` is sent even if API Key is empty.

derjust:
Thanks for the review

> Image and GIF in README needs to be absolute URL rather than relative.

✅
> Missing option names for Header/Query select, Color, and Method.

✅
> The colors "Danger", "Inverse", and "Transparent" seems to be the same as the Primary?

Removed those - they seemed to be not styled ✅

> Make the README heading the same as the title of the plugin, e.g. “Button panel”.

✅

---
>I’d expect a button like this to support adding custom headers and query parameters....
> By supporting generic headers and params, the user can still add secrets, but it’s at least not actively encouraged.
This would make it way more flexible; I agree with that.

> x-api-key is sent even if API Key is empty.
Let me check again. Couldn't reproduce this yesterday

derjust:
>I’d expect a button like this to support adding custom headers and query parameters....
By supporting generic headers and params, the user can still add secrets, but it’s at least not actively encouraged. 

Addressed also the last point  to have a dynamic list of parameters for query/parameters & release a new version of the plugin

marcusolsson:
Could you look at resolving these conflicts?

derjust:
> Could you look at resolving these conflicts?

Rebased & fixed conflict

marcusolsson:
Thank you!

PR #674, opened 2020-07-10
update discrete panel to 0.1.0

ryantxu:
The updates do two things:
* clean up the version.  dooh
* fix the field displayName settings in 7x


daniellee:
Thanks @ryantxu

PR #686, opened 2020-07-21
Update Multistat to 1.4.1 to fix bug with Grafana 7.1

michaeldmoore:
Grafana 7.1 removed support for the variableSvr injection parameter in the constructor, breaking this and presumably other panels built under the Angular model.

This release also updates the event mechanism (removing the named events warning) and adds a new 'sum' aggregation variant,a s requested by users.

**This is an urgent fix**

### DIFF
--- a/repo.json
+++ b/repo.json
@@ -3356,6 +3356,11 @@
           "version": "1.4.0",
           "commit": "20b6f3a04e49b32c8d3ca66796426eff83738f1b",
           "url": "https://github.com/michaeldmoore/michaeldmoore-multistat-panel"
+        },
+        {
+          "version": "1.4.1",
+          "commit": "eb7c20fe699a0cbb84a6e37a8039ec7bfca9afb8",
+          "url": "https://github.com/michaeldmoore/michaeldmoore-multistat-panel"
         }
       ]
     },


briangann:
is this 7.1+ or any other restrictions?

You can add a dependency option that will limit where it can be installed.

inside src/plugin.json like this, to specify for v7+

```
"dependencies": {
    "grafanaDependency": ">=7.0.0",
    "grafanaVersion": "7.x.x",
    "plugins": []
  }```

(the grafanaVersion is ignored when evaluating version upgrades)

briangann:
published, leaving open for dependency setting info :)

michaeldmoore:
Much appreciated.

I'll have yet another version to release sometime in the next couple of
weeks - I'll test out a number of back versions of Grafana and figure out
which ones matter, and if possible, figure out work arounds so the same
code base works with them all.  At least, back to \Grafana 6.0, if I can.

The Angular-to-React changes with Grafana 7.x seem to have been the most
significant.

I have a small request regarding this.  New versions of Grafana come out
quite frequently.  Frequently enough for those of us supporting independent
plugins don't have time to keep on track of each and every beta release.
To us, it seems, the official releases come 'out of the blue' and we
scramble to figure out if everything still works etc.  If you could create
an opt-in mailing list to send notifications shortly before an official
version release is made - it would give us an opportunity to check-out the
latest nightly build or whatever and see if there is anything we need to
fix - or report back to you - before the release actually happens.  I've
had to scramble to fix breaking changes due to Grafana releases several
times now.

That's all.  Thanks for your help and for creating and maintaining Grafana
itself.  Such a cool system
Michael Moore

marcusolsson:
@briangann @michaeldmoore Can this be closed?

michaeldmoore:
Yes.  Is it MY job, to close a PR?  I haven't had to do this before, but
maybe the procedure has changed.

marcusolsson:
No sorry, I was referring to this comment by @briangann.

> published, leaving open for dependency setting info :)

Was unsure whether he was waiting for more information :)

PR #711, opened 2020-08-25
Release Cognite data source v2.0.1

polomani:
Hi! We would like to release a patch for the recent major of our data source

#### Bug fixes:
- **\[Custom query\]** Option to display legend label as a plain text instead of expression

  Example for query: `ts{id=1} + ts{id=2} + 1`:

  * User label `custom name` will result to `custom name` label. (No specific timeseries referenced in the label)
  * User label `{{name}}` will result to `timeseries_name_1 + timeseries_name_2 + 1` label. (This works the same way as before)
- **\[Custom query\]** Error handling improved
  * HTTP 500 errors messages are visible from GUI
- **\[Custom query\]** Default granularity falls back to 1 second if shorter value is provided 
  * this used to happen when user chose time span less than 10 minutes, resulting with HTTP 400 error

### DIFF
--- a/repo.json
+++ b/repo.json
@@ -3569,6 +3569,11 @@
           "version": "2.0.0",
           "commit": "fbf4feeb10215a3ed04dd59d8f799bb5d293d707",
           "url": "https://github.com/cognitedata/cognite-grafana-datasource"
+        },
+        {
+          "version": "2.0.1",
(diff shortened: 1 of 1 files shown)


daniellee:
Merged and published

polomani:
@daniellee thank you!

PR #717, opened 2020-08-31
Add Timepicker Buttons Panel plugin

WilliamVenner:
Hey, back again after rewriting my plugin for Grafana 7.0.0.

### DIFF
--- a/repo.json
+++ b/repo.json
@@ -4423,8 +4423,8 @@
       "url": "https://github.com/WilliamVenner/grafana-timepicker-buttons",
       "versions": [
         {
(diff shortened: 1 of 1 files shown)


WilliamVenner:
I've included two versions in this addition to repo.json; one which works Grafana `<7.0.0`, and one that works with `>=7.0.0`. Is that the intended purpose, or shall I only add a single version?

marcusolsson:
Looks great! [Published](https://grafana.com/grafana/plugins/williamvenner-timepickerbuttons-panel)! 🎉

PR #730, opened 2020-09-09
Instana Grafana plugin version 3.0.1

Dalje-et:
- apiToken field is now of type password
- fixed custom label bug
- fixed page transition beacon bug


marcusolsson:
Published!

PR #774, opened 2020-10-26
Instana datasource plugin 3.1.1 and 2.7.5

schmidtl:
# 3.1.1
* fix granularity while adjusting time windows
* extend cache length
* backoff-retry for concurrent execution limit errors

# 2.7.5
* extend cache length
* backoff-retry for concurrent execution limit errors


marcusolsson:
Published! 🎉

PR #785, opened 2020-11-13
Remove BT plugins

jonasbleyl:
We are removing these plugins as we no longer maintain them and they are incompatible with the latest versions of Grafana.


marcusolsson:
We've removed them from grafana.com/plugins.

PR #797, opened 2020-11-23
Cognite datasource patch v2.1.1

polomani:
We encountered some issues with the latest release https://github.com/grafana/grafana-plugin-repository/pull/791

In version 2.1.0 `$variable` used as an `Asset tag` worked for earlier created dashboards, but it was impossible to create a new panel with the same capability. It was also not possible to see which variable is used as empty input field was shown instead.

Before the fix:
Screenshot 2020-11-23 at 10 06 46

After:
Screenshot 2020-11-23 at 10 06 30

### DIFF
--- a/repo.json
+++ b/repo.json
@@ -3585,6 +3585,17 @@
               "md5": "c661fa8d603deed7576c348d4bc8ef88"
             }
           }
+        },
+        {
+          "version": "2.1.1",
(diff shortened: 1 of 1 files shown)


marcusolsson:
[Published](https://grafana.com/grafana/plugins/cognitedata-datasource?version=2.1.1)! 🎉

polomani:
@marcusolsson thanks!

PR #798, opened 2020-11-23
Woutervh mapbox panel 1.0.0

woutervh-:
My first publish of a new plugin: `woutervh-mapbox-panel`.

Readme:
https://github.com/woutervh-/grafana-mapbox/blob/master/README.md

Release:
https://github.com/woutervh-/grafana-mapbox/releases/tag/1.0.0

### DIFF
--- a/repo.json
+++ b/repo.json
@@ -4719,6 +4719,24 @@
           }
         }
       ]
+    },
+    {
+      "id": "woutervh-mapbox-panel",
+      "type": "panel",
+      "url": "https://github.com/faxm0dem/grafana-riemann-websocket-datasource",
+      "versions": [
+        {
+          "version": "1.0.0",
+          "commit": "ca598f7ae61a95d48614e5c2d03e8497f33db738",
+          "url": "https://github.com/woutervh-/grafana-mapbox",
+          "download": {
+            "any": {
+              "url": "https://github.com/woutervh-/grafana-mapbox/releases/download/1.0.0/woutervh-mapbox-panel-1.0.0.zip",
+              "md5": "4fbd4415da58b2875d03ba0efc6d34e9"
+            }
+          }
+        }
+      ]
     }
   ]
 }
--- a/repo.json
+++ b/repo.json
@@ -4727,12 +4727,12 @@
       "versions": [
         {
           "version": "1.0.0",
-          "commit": "ca598f7ae61a95d48614e5c2d03e8497f33db738",
+          "commit": "a0f179e25bcc84d3745709d03fe86e3d604daa50",
           "url": "https://github.com/woutervh-/grafana-mapbox",
           "download": {
             "any": {
               "url": "https://github.com/woutervh-/grafana-mapbox/releases/download/1.0.0/woutervh-mapbox-panel-1.0.0.zip",
-              "md5": "4fbd4415da58b2875d03ba0efc6d34e9"
+              "md5": "d29054ffeba8ecad221cee9963323fc0"
             }
           }
         }
--- a/repo.json
+++ b/repo.json
@@ -4974,7 +4974,7 @@
     {
       "id": "woutervh-mapbox-panel",
       "type": "panel",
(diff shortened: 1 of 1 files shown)


marcusolsson:
Hi @woutervh-, and thanks for considering contributing your plugin! 

It seems like there's a few validation issues that you can see on the [plugin validator](https://github.com/woutervh-/grafana-mapbox/releases/download/1.0.0/woutervh-mapbox-panel-1.0.0.zip). Could you please take a look and resubmit once it passes the check.

woutervh-:
Hi @marcusolsson 

Thanks for the info. I didn't know about this validation app.

I've been able to fix all the errors and warnings it gives me, but I can't get this one fixed:
`Plugin has been modified since it was signed`

I've followed all the steps from the Package a plugin and Sign a plugin guides:
https://grafana.com/docs/grafana/latest/developers/plugins/package-a-plugin/
https://grafana.com/docs/grafana/latest/developers/plugins/sign-a-plugin/

And I'm using an API key with the PluginPublisher role from my Grafana account.

After running the build and sign commands, I don't change anything inside the `dist/` folder, so I don't know why it won't pass validation. Any clues what could be going on?

Thanks

WouterVanheeswijk-TomTom:
@marcusolsson 

Looking at the `MANIFEST.txt` file, I found that it generated backslashes in the file names:
image

I'm not a Go expert, but could this be an issue when running these file commands on your back-end?
https://github.com/grafana/plugin-validator/blob/103b8b0c0271203014cda1c5af21973697a7e6c4/pkg/plugin/signature.go#L58

woutervh-:
Ok I've managed to get it working by adding a replace inside the toolkit code.
image

Validator passes now!

image

woutervh-:
Hi @marcusolsson 

I've managed to get the manifest working through the validator. I had to work around a Windows path issue. Perhaps it's useful to mention this as a warning on the documentation?

Besides that, I've also updated the PR with new md5. Please check again.

Thanks!

marcusolsson:
Thank you for fixing the issues! 

This is ready to be reviewed 👍  I'm not able to give you an ETA, but feel free to check out the [review pipeline](https://github.com/grafana/grafana-plugin-repository/projects/2) for an indication of when we'll be able to review your plugin.

repo.json
```diff
@@ -4970,6 +4970,24 @@
           }
         }
       ]
+    },
+    {
+      "id": "woutervh-mapbox-panel",
+      "type": "panel",
+      "url": "https://github.com/faxm0dem/grafana-riemann-websocket-datasource",
```
marcusolsson, line 4977: Should this be https://github.com/woutervh-/grafana-mapbox?

marcusolsson:
Hi @woutervh-,

I've reviewed your plugin, and I'm happy to confirm it seems to work as expected!

My only concern is whether there's any licensing that prevents using Mapbox for commercial purposes etc? For example, in the documentation for Mapbox GL, it mentions an accessKey, which I can't find in your repository. I guess you're only using their package, not their API?

Are you allowed to freely develop and publish products built using Mapbox? Even if Grafana is open source, any plugin published on Grafana.com can be installed by our Grafana Cloud users, so it would likely be considered commercial use.

repo.json
woutervh-: Oops, you're right, I'll fix it.

woutervh-:
Good point @marcusolsson 

Luckily the version of mapbox-gl being used by the plugin is under the BSD-3 license, which allows commercial use:
https://choosealicense.com/licenses/bsd-3-clause/
(I say luckily, because as of mapbox-gl version 2, they have a more restrictive license.)

The only thing that I'm thinking about now, is that the BSD-3 license requires, that redistributions of the source code/compiled form, must also include their copyright notice. It looks like the webpack script created a bundle, which includes `module.js.map`, where the license is reproduced. I'm not a legal expert but I think that should be enough??

marcusolsson:
I agree. I've published your plugin, you should be able to install it with

`grafana-cli plugins install woutervh-mapbox-panel`

Thanks for contributing! ❤️

PR #799, opened 2020-11-26
Adding thalysantana-appcenter-datasource v1.0.0

thalysantana:
A User API Token is required to connect to the App Center. 
Install instructions provided on README.md. 
Plugin signed and packaged as required by Grafana's team.

### DIFF
--- a/repo.json
+++ b/repo.json
@@ -4748,12 +4748,12 @@
       "versions": [
         {
           "version": "1.0.0",
-          "commit": "875041934274c51effee10f0ad8609e32f8dd0d4",
+          "commit": "8cb0a9a0039267d571c329479fc911f02fe25d16",
           "url": "https://github.com/thalysantana/grafana-appcenter-datasource",
           "download": {
             "any": {
               "url": "https://github.com/thalysantana/grafana-appcenter-datasource/releases/download/v1.0.0/thalysantana-appcenter-datasource-1.0.0.zip",
-              "md5": "6d9ca2a81221cb1c770a6949cc3d61fd"
+              "md5": "7505a95c3a54dc82f887d5f2466648d3"
             }
           }
         }
--- a/repo.json
+++ b/repo.json
@@ -4748,12 +4748,12 @@
       "versions": [
         {
           "version": "1.0.0",
-          "commit": "8cb0a9a0039267d571c329479fc911f02fe25d16",
+          "commit": "3a1dcabd983467be6926ca6fa6e726a0a86d8e7d",
           "url": "https://github.com/thalysantana/grafana-appcenter-datasource",
           "download": {
             "any": {
               "url": "https://github.com/thalysantana/grafana-appcenter-datasource/releases/download/v1.0.0/thalysantana-appcenter-datasource-1.0.0.zip",
-              "md5": "7505a95c3a54dc82f887d5f2466648d3"
+              "md5": "c1064d6b2463bbff2588c2d9d51a32ba"
             }
           }
         }
--- a/repo.json
+++ b/repo.json
@@ -4740,6 +4740,24 @@
           }
         }
       ]
+    },
+    {
+      "id": "thalysantana-appcenter-datasource",
+      "type": "app",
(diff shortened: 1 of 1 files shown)


marcusolsson:
Looks great! 🎉 I'll add it to the review queue!

marcusolsson:
I've had the chance to review the plugin. I didn't know about App Center, but it looks really cool!

I first tried the plugin on 7.3.6, with limited success. It seems like the plugin doesn't yet support 7.3.x. 
Screen Shot 2021-01-29 at 15 07 59

I can see in your plugin.json that you only support 7.2.x (rather than >=7.2.0 or 7.x.x), so I if this is intentional I won't ask you to change at this point. In the README, it seems like you intend to support later versions as well. Please confirm what versions of Grafana you're targeting with this plugin.

I was able to get all query types except "apps" to work. When I selected "apps", the query seems to run indefinitely. Since the other queries works, I don't think it's related to my internet connection. Is there any other reason why this query doesn't complete?

Minor suggestions that I'll leave up to you to decide if you want to change:

- I would consider using a more user-friendly label for the query type, i.e. "Error groups" rather than "errorGroups"
- Grafana uses sentence case for headings and other labels. If you change "Query Type" to "Query type" it will be more consistent with the rest of Grafana

Finally, I'd recommend that you add to your documentation whether I need a Full Access token or Read Only token (I assumed I could use the Read Only one).

thalysantana:
Hi  Marcus.


   - Labels - I've updated them to follow the pattern you suggested.
   - Token access type - I've changed the readme file to mention that a
   Full Access token is required
   - Apps query - Please test it using a full access token. App Center
   requires it to run this query.
   - Image Error - I wasn't able to reproduce the error. Could you give
   more information about it?
   - Grafana 7.3.x issues - I've been testing using Grafana v7.3.7
   (1e261642f4) and haven't found any problems. Could you provide more
   information about the errors you got?


   - Grafana v7.3.7 (1e261642f4)

About the supported version, the Grafana Toolkit added the 7.2.x to the
plugin.json when I created the project, but I've been using it on 7.3.7
without any issues. How can I be sure it works on both versions?

I will wait for your answer so I can fix all the issues and push
the changes to Github.
Thanks for your time and patience to review this.

King regards,

Thálys Santana
Skype: thalysantana

marcusolsson:
I tested the plugin again, and it seems like the `meta` of null is due to the lack of a default query type. Once I select a query, it works.

Also, I now noticed that the plugin logs the instance settings. Could you remove this line?

https://github.com/thalysantana/grafana-appcenter-datasource/blob/875041934274c51effee10f0ad8609e32f8dd0d4/src/DataSource.ts#L26

The `grafanaDependency` property will be matched against the currently running Grafana version to determine whether the author supports that version or not. It doesn't guarantee that it will work. If you don't yet know what versions you support, I suggest that started with something like `>=7.0.0` which will let users install your plugin as long as they at least run 7.0.0.

thalysantana:
Hi Marcus.
I have made the modifications you suggested and built the release. I am
just not sure about how I should proceed.
Should I create a new merge request to grafana-plugin-repository, or just
fix the code on the current endpoint (
https://github.com/thalysantana/grafana-appcenter-datasource/releases/download/v1.0.0/thalysantana-appcenter-datasource-1.0.0.zip)?


Kind regards,

Thálys Santana
Skype: thalysantana

marcusolsson:
You can keep this PR, and the URL if you'd like, but please update the commit and the hash for the new release 🙏

marcusolsson:
Changes looks good, but now there seems to be a missing screenshot. If it has been removed, it should be removed from the plugin.json as well.

CleanShot 2021-02-11 at 15 56 19@2x

thalysantana:
Hi Marcus,
I've fixed the issues with the screenshots.

Kind regards

Thálys Santana
Skype: thalysantana

marcusolsson:
Published! 🎉

thalysantana:
Thanks a lot. 🙌

PR #828, opened 2021-01-08
Add marcusolsson-gantt-panel

marcusolsson:


leeoniya:
@marcusolsson let's coordinte before you put too much additional work into this ;)

see https://github.com/grafana/grafana/issues/30129

marcusolsson:
Sure! Though I'm not sure I see the overlap (assuming that's what you meant)?

leeoniya:
heh, i'm on the other side, trying to find the distinction 🤣, besides this plugin (which is a gantt minus the time header table and task tree?) being "in the future" (scheduled), and the other being in the past. if you have plans to add the task tree to this and other gantt features, then i agree that this would be sufficiently distinct.

marcusolsson:
I have limited experience using the discrete panel, but my understanding is that it visualizes state changes in a long running process.

With the Gantt panel, I wanted to focus more on batch jobs rather than time series. For example, comparing multiple runs of a multi-stage pipeline that runs every night (much like the graphs for distributing tracing). The main use case have been to visualize DAG runs in Apache Airflow:

download

Depending on user feedback, I don't mind adding more Gantt features like completion (partially filled activities) and dependencies, etc.

My main concern would be that a "bars over time" panel would become to generic to be useful (maybe better suited as a component?), but I'd be happy to keep the discussion going!

leeoniya:
the goal for the new Timeline panel will be something very similar to what you're showing (tracking state over time).

it's probably too soon to dig into the details, and you're right that it may be too generic if you plan to add more gantt-specific features. it may be possible to extend the Timeline plot with the additional stuff you need on top of it. just wanted to give you a heads up so we're not duplicating the same work unless it's warranted.

cc @ryantxu 

e.g.: https://grafana.com/grafana/plugins/natel-discrete-panel

and:

image

ryantxu:
I don't see much problem in having overlap from marcusolsson-* panels -- these tend to fill in gaps, evaluate options, serve as templates people can use to customize etc.  Looking across the whole plugin ecosystem, there will always be many ways to slice your problems.  With the grafana core panels, we do need to worry about differentiation, maintenance etc.

leeoniya:
k, sounds good!

PR #833, opened 2021-01-22
add rockset backend plugin

pmenglund:
I'm not sure this PR is valid - the plugin is open source, but the backend if closed source (with a free tier)

### DIFF
--- a/repo.json
+++ b/repo.json
@@ -5192,6 +5192,24 @@
           }
         }
       ]
+    },
+    {
+      "id": "rockset-backend-datasource",
+      "type": "datasource",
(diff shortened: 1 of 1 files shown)


marcusolsson:
For plugins that only work with commercial offerings, you'll need a plugin subscription. For more information on commercial plugins, refer to https://grafana.com/legal/terms/.

Let me know if this is interesting to you!

pmenglund:
We only have a handful of customers who use our grafana plugin, so we'll need a larger audience before it makes commercially sense to get a plugin subscription as I assume it runs in the tens of thousands of USD.

PR #844, opened 2021-01-28
Update Multistat to v1.7.0

michaeldmoore:
This version adds label and group renaming, plus reformatting date/time field in-place

Released using new github workflow

### DIFF
--- a/repo.json
+++ b/repo.json
@@ -3470,6 +3470,66 @@
       "url": "https://github.com/michaeldmoore/michaeldmoore-multistat-panel",
       "versions": [
         {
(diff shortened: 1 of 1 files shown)


marcusolsson:
I've published the plugin, but I'd appreciate if you could add back the previous versions in the PR, as I assume you don't mean to unpublish them 🙏

michaeldmoore:
Thanks and will do...

michaeldmoore:
I re-added the old Multistat version references, as requested.

The new commit hash is: ad8a46dd6d6211dcffb55b534dd596f6b4879276

PR #850, opened 2021-02-02
Instana datasource plugin 3.1.4

schmidtl:
* fix query range on time window or rollup adjustments 
* display key input box for empty key value filter


marcusolsson:
Published! 🎉

PR #863, opened 2021-02-09
add auxmoney-waterfall-panel 1.0.3

cawolf:


marcusolsson:
Hi @cawolf,

This plugin has a `private` signature. To publish on grafana.com, it needs to be signed as Community or Commercial. Would you like me to configure a Community signature for this plugin? You will have to sign the plugin again and update your PR with the new signature.

cawolf:
Hi @marcusolsson ,
yes, we would like to use a community signature. The corresponding ticket was declined from your support, and instead a direct PR to this repository was proposed.
We will sign the plugin again as soon as we get the new signature from you (you can use our plugin contact email address opensource@auxmoney.com).

cawolf:
Hi @marcusolsson , 
did you provide the community signature to us already? If so, how can we retrieve it?

marcusolsson:
I've configured `auxmoney-waterfall-panel` to be signed as a Community plugin. Please make sure that you use a API key that was created using the `auxmoney` org account when you sign it.

marcusolsson:
I've had the chance to review the plugin:

- The plugin selects the field to use based on the field name. Since the plugin requires three separate types of fields, I recommend that you instead look for fields of a specific type. For example, instead of 

  ```
  series.fields.filter(field => field.name === 'Time')[0];
  series.fields.filter(field => field.name === 'value')[0];
  series.fields.filter(field => field.name === 'name')[0];
  ```

  ```
  series.fields.find(field => field.type === FieldType.time);
  series.fields.find(field => field.type === FieldType.number);
  series.fields.find(field => field.type === FieldType.string);
  ```

  I find that it makes it easier to get started. Alternatively, let the users select the name of the field to use.

- I'd recommend that you add some docs in the README on how to configure the panel. Such as what fields does the panel expect (name/type), which ones are required/optional, and what are they used for. I had to look at the source code to figure out what it expected.

- The label is currently drawn outside of the bar. Maybe add some sensible padding?

  Screenshot 2021-03-22 at 15 25 42@2x

- Finally—and considered this more of a feature request—I think it would've been helpful to have some form of relative/absolute time on the x-axis to help me see where when the events happened, rather than just their duration.

I don't think any of these points are blockers, but I'd recommend making it a little easier to get started by addressing my first two points. Let me know if you'd like to address any of the feedback 👍

cawolf:
I built a new version and implemented the 2 main changes you proposed. I will create separate issues for your feature request and the padding / UI issues soon.

marcusolsson:
v1.0.4 looks ready to publish, but v1.0.3 doesn't have a valid signature. Could you remove 1.0.3 from the PR and I'll publish this?

cawolf:
Hey @marcusolsson , I removed the invalid signed version as you requested.

marcusolsson:
I've published the plugin to the marketplace 🎉 , but please allow for a few minutes before it's available.

marcusolsson:
Looks like you may have forgotten to update the package.json version. I published 1.0.4, but it's listed as 1.0.3 on the website. It doesn't matter so much to us, but we should probably update repo.json to reflect this.

cawolf:
You are right, there is a release script misconfiguration. I will sort this out and create a new PR for this.

Thanks for taking your time and publishing our pluging! :+1:

PR #592, opened 2020-04-05
added geeksrus-mqtt-panel V1.0.7

geeks-r-us:

PR #866, opened 2021-02-10
Added geeksrus-mqtt-panel 1.0.7

geeks-r-us:
Next try of submitting geeksrus-mqtt-panel previous was #592

This is a panel that allows to connect to mqtt servers through ws and wss and adds control elements.

### DIFF
--- a/repo.json
+++ b/repo.json
@@ -5540,6 +5540,24 @@
           }
         }
       ]
+    },
+    {
+      "id": "geeksrus-mqtt-panel",
+      "type": "panel",
+      "url": "https://github.com/geeks-r-us/mqtt-panel",
+      "versions": [
+        {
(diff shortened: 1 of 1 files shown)


marcusolsson:
Hi @geeks-r-us,

This plugin stores secrets as panel options, which poses a critical security vulnerability. Anyone with Viewer permissions will have access to the information. This means that even anonymous users will be able to see the password in cleartext. 

Because of this, I won't be able to publish this plugin in its current state. 

I **strongly** recommend that you document this for users in your README as well in the panel options so that users are aware of the risks when using this plugin.

Let me know if you have any questions.

marcusolsson:
Closing this for now, but please let me know if you'd like to address the feedback and I'll reopen this.

PR #13, opened 2016-04-14
Add the OpenNMS datasource to the list of available plugins.

j-white:
Great job on the plugin architecture! I'm happy I don't have to maintain my own RPMs anymore :)

PR #10, opened 2016-04-13
alexanderzobnin-zabbix-app version 3.0.0-beta6

alexanderzobnin:

PR #867, opened 2021-02-11
Update gapit-htmlgraphics-panel to v1.3.1

ZuperZee:
Update v1.3.0 for [gapit-htmlgraphics-panel](https://github.com/gapitio/gapit-htmlgraphics-panel/tree/main)

This update fixes an issue where the panel options didn't update when importing/changing values in the "Import/Export" option.
onpanelupdate events has been added to make it easier to integrate React and Svelte into the panel 
Here is the [docs](https://github.com/gapitio/gapit-htmlgraphics-panel/tree/v1.3.0#react) and an [example](https://github.com/gapitio/htmlgraphics-react-radar)(example is a little outdated, but the general is there) with react.
If you use the example in grafana remember to add some metrics (inside-monday -> inside-friday) and (outside-monday -> outside-friday)

Also a huge update to the file naming and the code. A lot of the code has been made simplier and more organized.

## v1.3.0 (2021-02-11)

- **onInit**: Add onpanelupdate events, dynamicData option, and "Trigger panelupdate when mounted" option [#10](https://github.com/gapitio/gapit-htmlgraphics-panel/pull/10)

### Bug fixes

- **Import/Export**: Fix panel options not updating when importing/changing values in the import option [#13](https://github.com/gapitio/gapit-htmlgraphics-panel/pull/13)

### DIFF
--- a/repo.json
+++ b/repo.json
@@ -5467,6 +5467,17 @@
               "md5": "546f08335ccf4f73e38b1d669dc0687e"
             }
           }
+        },
+        {
+          "version": "1.3.0",
+          "commit": "777d3f936ee1d2e926e29e701d3b8a788e12542a",
+          "url": "https://github.com/gapitio/gapit-htmlgraphics-panel",
+          "download": {
+            "any": {
+              "url": "https://github.com/gapitio/gapit-htmlgraphics-panel/releases/download/v1.3.0/gapit-htmlgraphics-panel-1.3.0.zip",
+              "md5": "fd9e8a320a5605ee52df4c9301e34950"
+            }
+          }
         }
       ]
     },
--- a/repo.json
+++ b/repo.json
@@ -5469,13 +5469,13 @@
           }
         },
         {
-          "version": "1.3.0",
-          "commit": "777d3f936ee1d2e926e29e701d3b8a788e12542a",
+          "version": "1.3.1",
+          "commit": "385da3cf551631e4c16d3c045e2da420129f13a0",
           "url": "https://github.com/gapitio/gapit-htmlgraphics-panel",
           "download": {
             "any": {
-              "url": "https://github.com/gapitio/gapit-htmlgraphics-panel/releases/download/v1.3.0/gapit-htmlgraphics-panel-1.3.0.zip",
-              "md5": "fd9e8a320a5605ee52df4c9301e34950"
+              "url": "https://github.com/gapitio/gapit-htmlgraphics-panel/releases/download/v1.3.1/gapit-htmlgraphics-panel-1.3.1.zip",
+              "md5": "2e650d2ced1a4811f0f030474e9b68b5"
             }
           }
         }


ZuperZee:
Does the https://github.com/gapitio/gapit-htmlgraphics-panel/blob/777d3f936ee1d2e926e29e701d3b8a788e12542a/src/plugin.json#L49 mean that it can only be installed on v7.4.x?
What decides what version it can run on?
How can I make sure it runs on as many versions as possible while being stable?

marcusolsson:
The `grafanaVersion` property is only used for display purpose and isn't actually checked of enforced. It has been deprecated in favor of `grafanaDependency` which must be a valid semver string. 

AFAIK, the only place `grafanaDependency` is used at the moment is in grafana-cli, where the currently running Grafana version is matched to see if the plugin supports it. In your case, since it's set to `>=7.4.0` only users with at least 7.4 can install your plugin using grafana-cli. 

Mostly, it depends on what features your using from the `@grafana/*` packages. If you try to use a function that is only available in later versions, you'll get an error when trying to load it. If you only use features available in 7.0.0, your plugin should be able to run all the way back.

Your next question then will probably be: "How will I know when a certain feature was introduced?", which unfortunately means you'll either have to check the CHANGELOG, or (arguably the easier option) to simply test your plugin in earlier versions.

ZuperZee:
Would it be better to remove grafanaVersion in favor of grafanaDependency?

marcusolsson:
The plugin marketplace still uses the `grafanaVersion`, so if you remove it, users won't see the requirement unfortunately.

With `grafanaVersion`:

CleanShot 2021-02-12 at 10 59 35@2x

Without `grafanaVersion`:

CleanShot 2021-02-12 at 11 00 19@2x

Other than that, there's no harm in leaving it there, or removing it.

ZuperZee:
Ok, thank you. Sounds like leaving is better.

marcusolsson:
I'm happy to publish this right away, but I just want to confirm that you wish to update the minimum required Grafana version to 7.4.0? Are you using features that were introduced in 7.4?

ZuperZee:
> I'm happy to publish this right away, but I just want to confirm that you wish to update the minimum required Grafana version to 7.4.0? Are you using features that were introduced in 7.4?

I tested with 7.0.0 and it seems to complain about prettier and missing props, but except that it works. Is there a changelog for all plugin related changes?

marcusolsson:
If the end user is able to run the plugin without problem on 7.0.0, I'd recommend that you keep the `grafanaDependency` as `>=7.0.0`. Otherwise, you'll be unnecessarily restricting who'll be able to run your plugin (the majority of users have yet to upgrade to 7.4).

marcusolsson:
`grafanaDependency` does **not** have to match the version of grafana-toolkit that the plugin is using.

ZuperZee:
Ok, I'll test some more and I'll commit the new changes later today

ZuperZee:
Latest working version was 7.1.0 since that's when context came out.
I didn't find it mentioned anywhere in the changelog on grafana/grafana.

marcusolsson:
Published! 🎉

PR #880, opened 2021-02-24
Add StableNet® Data Source

fafeitsch:
This PR adds a datasource for displaying  [StableNet®](https://www.infosim.net/stablenet/) Data in Grafana®

We are happy to provide access to a testing StableNet® environment. Please contact us via private message or E-Mail.

The plugin is not yet signed because it's the initial release and according to the docs (https://grafana.com/docs/grafana/latest/developers/plugins/sign-a-plugin/) it first must be reviewed.

### DIFF
--- a/repo.json
+++ b/repo.json
@@ -2234,7 +2234,6 @@
             }
           }
         }
-
       ]
     },
     {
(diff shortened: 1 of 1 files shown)


marcusolsson:
It seems like StableNet is a commercial product, and I can't find any open source version available. To publish plugins for commercial use, you need a plugin subscription. Are you affiliated with StableNet, and if so, is this something that would be interesting to you?

fafeitsch:
Thanks @marcusolsson for your reply. It seems that we've misinterpreted the requirements of plublic plugins. We assumed that only the plugin itself must be open-source, and not the whole program it's connected to.

We are already evaluating other options, including a plugin subscription.

PR #884, opened 2021-03-01
Add PingCAP customized loki datasource plugin

baurine:
This PR is to add a PingCAP customized loki datasource plugin. The plugin enhances the experience to use loki explore page, enable to select many variables on the explore page.

*the screenshot comes from dev env*

explore

### DIFF
--- a/repo.json
+++ b/repo.json
@@ -2234,7 +2234,6 @@
             }
           }
         }
-
       ]
     },
     {
@@ -5660,6 +5659,24 @@
             "any": {
               "url": "https://github.com/thalysantana/grafana-appcenter-datasource/releases/download/v1.0.0/thalysantana-appcenter-datasource-1.0.0.zip",
               "md5": "c1064d6b2463bbff2588c2d9d51a32ba"
+            }
+          }
+        }
+      ]
+    },
+    {
+      "id": "customized-loki-datasource",
+      "type": "datasource",
+      "url": "https://github.com/baurine/customized-loki-ds-plugin",
+      "versions": [
+        {
+          "version": "0.0.9",
+          "commit": "7822f06d01bd1d6591d272e4f82d277b5dbc0975",
+          "url": "https://github.com/baurine/customized-loki-ds-plugin",
+          "download": {
+            "any": {
+              "url": "https://github.com/baurine/customized-loki-ds-plugin/releases/download/v0.0.9/pingcap-customized-loki-0.0.9.zip",
+              "md5": "807e27297f990e5422128c99f62804c8"
             }
           }
         }


baurine:
Hi @marcusolsson , is there anything extra should I add for this PR?

marcusolsson:
No, it looks like it has the correct metadata and is ready to be reviewed. You can see your place in the review queue in the [project board](https://github.com/grafana/grafana-plugin-repository/projects/2).

baurine:
ping

baurine:
Hi @marcusolsson , lately we upgrade the grafana to 8.0 beta, and it doesn't support un-signed the plugin anymore, can you help accelerate the review for our plugin, thank you!

marcusolsson:
Could you share some more details on the background on this plugin? The docs mentions it's a customized data source of Loki for PingCAP. There are some mentions of TiDB in the source code. Is this plugin only intended to be used with PingCAP/TiDB?

If so, how can I set up a test environment with data that I can use to review the plugin? To accelerate the review of the plugin, could you please provide a Docker Compose file that configures a environment that we can use to test the plugin in?

baurine:
Hi @marcusolsson , yep, this is not a universal plugin, it is only can be used internally. 

This plugin is similar with the new `Log browser` feature of loki, but it is created before `Log browser`.

image

Our plugin improves the original loki plugin experience when it doesn't have the new `Log browser` feature.

image

At first, it will load all tenants; after user selects a tenant, then it will load all clusters of that tenant; after user selects a cluster, then it will load all instances of that cluster. Finally, it will auto generate the LogQL query string. Similar behavior with the new `Log browser` feature.

Tenants/clusters/instances, these information comes from prometheus, not loki self, because we don't store the tenant value in the loki. That's why this plugin depends on prometheus and loki both.

Not sure whether I explain it clearly.

If this kind of plugin isn't accepted to publish to the store, it is fine to close this PR. 

We just want to dismiss the `unsigned` warning, especially after upgrading to grafana 8.0, the unsigned plugin can't be loaded defaultly.

Not sure do we have some other ways to signature the plugin?

marcusolsson:
Gotcha! It sounds like this plugin is only going to be used internally at PingCap, and won't be useful for other users. It sounds like you'd be better off with a Private signature. For private signatures, you don't need to submit the plugin for review.

To create a Private signature, you need to specify the root URL, `--rootUrls`, of the Grafana instance you intend to run the plugin on, e.g. https://grafana.pingcap.com.

For more information, refer to the docs on how to [Sign a private plugin](https://grafana.com/docs/grafana/latest/developers/plugins/sign-a-plugin/#sign-a-private-plugin).

Let me know if you have any questions on signing private plugins!

baurine:
Great, let me have a try, thank you!

PR #885, opened 2021-03-01
Add anodot-panel plugin

gooduling:
Please use anodot-datasource and anodot-panel plugins together. 
**For datasource Auth configuration:**
1. Set https://bc-test.anodot.com/ as API URL
2. Set 0bb22f37ef90428f5af56536aa6ca94d as API Token
Screenshot 2021-03-01 at 18 14 33


marcusolsson:
The error seems to be related to incorrect JSON in the pull request. The plugins looks ready to be reviewed though, so I'm moving this along to Waiting for Review.

marcusolsson:
Hi @gooduling,

I'm testing the data source at the moment, but I can't seem to build a query that returns any data. Do you have an example of a query that I can use? Preferably one that I can use to test the panel as well.

I'm getting Internal Server Error for the Anomalies and Topology Map scenarios. Any idea why?

Also, I'm getting a couple of errors every now and then, though I haven't been able to consistently reproduce them. Maybe you have an idea?:

Screenshot 2021-03-16 at 09 24 40@2x

Screenshot 2021-03-16 at 09 24 07@2x

gooduling:
Hi @marcusolsson. I add 2 screenshots I've done right now for successful requests of Anomaly and Topology. Is it the same way you build the queries?
Screenshot 2021-03-16 at 10 51 05
Screenshot 2021-03-16 at 10 49 38

marcusolsson:
Thank you. The queries you provided seem to work for me 🙏

gooduling:
@marcusolsson thank you too. Please let me know if you have errors. I definitely should reproduce and handle them

marcusolsson:
I found a few feedback items on the panel. No blockers, but I think they would improve the plugin.

- I've set the time interval to one year (1), but it seems like the X-axis is fixed as it gets cut off (2) rather than adjusting to the selected interval. Indeed it seems like only the two first months (3) are visible. This was a little confusing at first, since I thought I was looking at the data from the full year.

   Screenshot 2021-03-16 at 10 43 50@2x

- Also, when I happened to resize the panel, I noticed that there were more graphs available if I scrolled. I think you should consider making it more clear that the query returned multiple results.

- I noticed that you're using Highcharts, which I believe requires a license for commercial purposes. I can't see any license in the plugin however. Would you might assuring me that we'd be able to publish the plugin using Highcharts for commercial purposes?

gooduling:
@marcusolsson thanks, good point about the timescales. Right now it draw only that period the backend gave the data. We probably should synchronize it all. As for Hicharts license, I'll come back with the answer soon

gooduling:
Hi @marcusolsson , the PR is updated

marcusolsson:
Thank you! It looks like the data source hasn't been signed yet however? Did you have any issues signing it?

Also, I see that you've set the grafanaDependency to 7.0.x. This means that only users of 7.0 can use your plugin, which is probably not what you intended. I suggest that you update it to `>=7.0.0`.

gooduling:
Hi Marcus,
It’s probably some misunderstanding from my side. I was sure it’s Grafana
who signs the plugin. I’ve updated the PR yesterday and was waiting for
your move. What exactly should I do to have the plugin signed?
(Except updating Grafana dependency, thank you.)

ср, 14 квіт. 2021 о 14:26 Marcus Olsson ***@***.***> пише:

marcusolsson:
You can find the instructions on how to sign the plugin here: https://grafana.com/docs/grafana/latest/developers/plugins/sign-a-plugin/

I've configured the plugin to be signed under the commercial signature level.

gooduling:
Hi @marcusolsson, I can't get what should I specify in rootsUrl param and I can't ignore it: **npx @grafana/toolkit plugin:sign --rootUrls https://example.com/grafana** . We want our plugin was available everywhere, wherever people has their Grafana installed. No any restrictions

marcusolsson:
rootUrls are only for privately signed plugins. You shouldn't need to set any flags.

The following should work:

```
export GRAFANA_API_KEY=<YOUR_API_KEY>
npx @grafana/toolkit plugin:sign
```

If the toolkit tells you to use rootUrls, that means that it couldn't find a signature level for the plugin. I've confirmed that they're both are configured as commercial. 

The other reason could be that you're using an API key that was generated from you personal account. The API key needs to be generated from the organization that owns the plugin, i.e. `anodot`.

gooduling:
@marcusolsson I think I'm getting API key from Anodot account (log in by andrii.ulianenko@anodot.com) and true, it still requires rootUrl
Screenshot 2021-04-15 at 10 10 25
Screenshot 2021-04-15 at 10 12 59

gooduling:
@marcusolsson Thank you. Was able to sign in plugins and have MANIFEST.TXT in both of them. Should I update this PR now wis new zips and md5 labels?

marcusolsson:
Awesome! Yes, please update the pull request with the new ZIPs and checksums, and I'll publish them right away

gooduling:
@marcusolsson Done! 🤞

gooduling:
@marcusolsson I'm sorry, seems like MANIFEST.txtx in anodot-datasource does not include backend-files like "gpx_anodot-datasource_darwin_amd64" or "gpx_anodot-datasource_linux_amd64" etc. Should it include also? With this files zip file is about 80mb size. I probably built the dist folder in a wrong way, without "mage build"

marcusolsson:
Yes, you need to run the sign command once you've built all the files that go into the final package. Every file needs to be accounted for in the MANIFEST.txt. For example, I can see that the index.html isn't included in the manifest either.

gooduling:
@marcusolsson have rebuilt and repacked all of it. Please check

marcusolsson:
Published! 🎉

gooduling:
Hi @marcusolsson 
There is an issue with the installation of our plugins with grafana-cli. It installs anodot-datasorce, then anodot-plugin, then anodot-datasource etc infinite times. https://github.com/anodot/grafana-panel/issues/1
I think now the reason is that each of our two plugins has specified the second one as a dependency in their plugin.json. Was it incorrect? Should we not mention them and let them be downloaded separately? If yes, how should I apply changes? Should I reopen this PR or create a new one? Thank you
Screenshot 2021-04-18 at 09 22 38

PR #908, opened 2021-03-26
Update frser-sqlite-datasource to v1.0.3 / v1.1.0 / v1.2.0

fr-ser:
This is an update to the existing plugin. v1.0.2 is skipped since it was released only shortly before v1.0.3

A summary of changes can be found here: https://github.com/fr-ser/grafana-sqlite-datasource/blob/master/CHANGELOG.md

### DIFF
--- a/repo.json
+++ b/repo.json
@@ -2854,6 +2854,36 @@
               "md5": "776aaf3fbfe3dd9f532dbd1a5e80981a"
             }
           }
+        },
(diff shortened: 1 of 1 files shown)


fr-ser:
I added v1.1.0 in addition to v1.0.3

marcusolsson:
Published! 🎉

PR #909, opened 2021-03-28
Release dalvany-image-panel 2.3.0

Dalvany:
Fix tooltip and allow to use variables for some options

Please note that I didn't sign my plugin since I don't yet have the signing level and the release is therefore still a draft.


Dalvany:
I made the release public but I can't rerun checks. Should I close this PR and open a new one ?

PR #912, opened 2021-04-03
Update repo.json

projects-pandorafm:
Hello from Pandora FMS, 
We want to upload our datasourse plugin to your library, so we can sign it and continue using it on Grafana 7.0+

Thanks in advance,
Kind regards

### DIFF
--- a/repo.json
+++ b/repo.json
@@ -5921,6 +5921,23 @@
           }
         }
       ]
+    },
+    {
+      "id": "pandorafms-datasource",
+      "type": "datasource",
+      "url": "https://github.com/projects-pandorafms/pandorafms-datasource",
+      "versions": [
+        {
+          "version": "1.0.1",
+          "url": "https://github.com/projects-pandorafms/pandorafms-datasource",
+          "download": {
+            "any": {
+              "url": "https://github.com/projects-pandorafms/pandorafms-datasource/releases/download/1.0/pandorafms-datasource.zip",
+              "md5": "ea390f4414b784966d8337680afc391a"
+            }
+          }
+        }
+      ]
     }
   ]
 }


marcusolsson:
Hi,

I've started reviewing the plugin using the Docker Compose in the repository. Unfortunately, it seems like the Grafana extension isn't included. What's the best way to enable it?

Screenshot 2021-07-06 at 14 56 08@2x

Screenshot 2021-07-06 at 14 56 55@2x

Also, I can't seem to find a Grafana Cloud account called `pandorafms`. To claim ownership of (and to sign) the plugin, you'll need to create a (free) Grafana Cloud account to publish plugins. The id of the account needs to match the first part of the plugin ID, i.e. `pandorafms`.

Let me know if you have any questions!

rafaelameijeiras:
Hello, 

you are right, didn't created the account before, thanks we already have an account with pandorafms name.

Regarding the testing whit docker env we tried in a couple of different machines on different networks and works, is it possible the pandora env wasn't up yet? It takes a minute to be ready after compose is up.

I made a little video with all the process we do to test the data source, maybe is helpful to identify the issue.

The Pandora url is for the docker compose internal network so is not possible to access from local or remote browser the name resolution just works inside the containers, in case you want to test the url you should use 8080 port which is exposed to outside host. EX: http://127.0.0.1:8080/pandora_console/extensions/grafana/

if works and Pandora is up, you will see an error like:  
image

Which means you are ready to connect but of course from the browser you don't send any authentication so, you are not allowed to retrieve any data but the datasource is ready.

So I will leave over here the video, I hope you find it useful: https://www.youtube.com/watch?v=LpSTlN2iWFU

thanks in advance

marcusolsson:
Hi @rafaelameijeiras,

We're deprecating this repository in favor of the new submission process. Could you please submit your plugin from your Grafana Cloud account? More information is available in the README for this repository.

PR #916, opened 2021-04-06
Update larona-epict-panel to v2.0.3

LucasArona:
Update larona-epict-panel to v2.0.3

### DIFF
--- a/repo.json
+++ b/repo.json
@@ -3657,12 +3657,12 @@
           }
         },
         {
-          "version": "2.0.2",
+          "version": "2.0.3",
           "url": "https://github.com/LucasArona/larona-epict-panel",
           "download": {
             "any": {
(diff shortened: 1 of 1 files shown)


marcusolsson:
Hi! I've reviewed the plugin and noticed that images aren't resized to fit the panel. Meaning, if the image is larger than the panel, it overflows and prevents the user from resizing the panel.

Screenshot 2021-05-10 at 15 47 59@2x

Is it possible to add an option to decide how to fit the image? Maybe let the user configure the [object-fit](https://www.w3schools.com/css/css3_object-fit.asp) (or similar)? 

I assume this is an existing issue so I won't block the release for it. Let me know if you'd like to fix this now, or wait until next release.

LucasArona:
OK, thanks for the heads up! Indeed, this was not introduced with this release.
I'll fix that in the next release, you can approve this one because it fixes other bugs and I prefer that users get these fixes quickly.

Best,
Lucas

marcusolsson:
Published! 🎉

PR #927, opened 2021-04-23
Slope graph panel

KatrinaTurner:
a new panel plugin for a slope graph.

### DIFF
--- a/repo.json
+++ b/repo.json
@@ -6021,6 +6021,23 @@
           }
         }
       ]
+    },
+    {
+      "id": "netsage-slope-graph-panel",
+      "type": "panel",
+      "url": "https://github.com/netsage-project/react-slope-graph",
+      "versions": [
+        {
+          "version": "0.0.1",
+          "url": "https://github.com/netsage-project/react-slope-graph",
+          "download": {
+            "any": {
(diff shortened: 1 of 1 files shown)


KatrinaTurner:
my check is failing because it can't find the README.md, but its definitely there.

repo.json
```diff
@@ -6021,6 +6021,23 @@
           }
         }
       ]
+    },
+    {
+      "id": "netsage-slope-graph-panel",
```
briangann, line 6026: the linter won't accept this id, it would need to be renamed to "netsage-slopegraph-panel"

briangann: also you'll want to run the zip file through this: https://plugin-validator.grafana.net/

I did a quick check and it says the zip file doesn't contain a plugin.json, but the zip does have one under "src".

the zip package is expecting everything to be prebuilt and "live" at the top level of the extracted zip file, with "module.js" file and everything else needed.

briangann:
i see what the problem is - the circleci config in your repo is not enough to get a plugin ready for publishing,  it would be better to switch to the github actions method and disable circleci.

KatrinaTurner:
Ah, I see. Okay, thank you so much! 😄

PR #937, opened 2021-05-05
riemann datasource plugin 0.1.5

faxm0dem:
Add support for events with numeric time (in seconds)
This makes this plugin also compatible with Mirabelle
https://github.com/mcorbin/mirabelle

### DIFF
--- a/repo.json
+++ b/repo.json
@@ -93,6 +93,24 @@
               "md5": "52f4d2f2a49451cc02832c35521ce2ef"
             }
           }
+        },
+        {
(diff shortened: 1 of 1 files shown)


marcusolsson:
LGTM! Though it seems like the PR format is a bit weird. Looks like the added version is an entire plugin object? Could you update this?

faxm0dem:
Sorry about that, this is fixed.
You can squash the 2 commits of course if you liks

marcusolsson:
Published! 🎉

PR #957, opened 2021-05-25
Cognite data source v2.3.0

polomani:
Hi,

we'd like to release a new feature for the Cognite plugin.

We added support for "nested/chained" variables. So one variable can be used as a filter in another variable expression.
In our case, if we have a hierarchy of assets, we can leverage this feature to allow the user to switch between assets on each level of the hierarchy.

Screenshot 2021-05-25 at 17 22 44
Screenshot 2021-05-25 at 17 22 55

I can provide the `.json` of this example dashboard for testing purposes (via email).

### DIFF
--- a/repo.json
+++ b/repo.json
@@ -4205,6 +4205,17 @@
               "md5": "3d744164de2cf59266be0777a9395afb"
             }
           }
+        },
+        {
+          "version": "2.3.0",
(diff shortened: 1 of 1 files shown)


marcusolsson:
Published! 🎉

polomani:
Thanks @marcusolsson !

PR #960, opened 2021-05-27
Instana datasource plugin 3.3.0

schmidtl:
* fix display of error messages
* security fix

### DIFF
--- a/repo.json
+++ b/repo.json
@@ -3223,6 +3223,17 @@
               "md5": "961cf44ff5ee92d25f4c723a5c1d332b"
             }
           }
+        },
+        {
+          "version": "3.3.0",
+          "commit": "6631df7f04241e69b8503e28a4ceff06de2eed5b",
+          "url": "https://github.com/instana/instana-grafana-datasource",
+          "download": {
+            "any": {
+              "url": "https://github.com/instana/instana-grafana-datasource/releases/download/v3.3.0/instana-datasource-3.3.0.zip",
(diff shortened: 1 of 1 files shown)


daniellee:
Testing with Grafana 8.0 Beta 2 and there is a problem with the CSS on the data source configuration page:

image

I will check if we have broken something in the Checkbox control and if it should be fixed before the Grafana release on June 8th. Seems to be something to do with the height.

Otherwise everything looks fine.

Dalje-et:
Is there anything we can do on our side to fix this?

daniellee:
No need - fixed in Grafana instead: https://github.com/grafana/grafana/pull/35022 It was a bug we introduced when working on the Grafana themes and broke the vertical layout.

daniellee:
@Dalje-et one tip I got from our frontend engineers is to wrap the `Checkbox` components with a `Field` component to get the right margins around each checkbox.

For example: https://github.com/grafana/grafana/blob/fix-checkbox-height/public/app/features/dashboard/components/LinksSettings/LinkSettingsEdit.tsx#L118-L141

daniellee (approved):
LGTM

Dalje-et:
> @Dalje-et one tip I got from our frontend engineers is to wrap the Checkbox components with a Field component to get the right margins around each checkbox.

Thanks a lot @daniellee. Will keep this in mind when doing another update of the datasource.

PR #963, opened 2021-06-02
Adds pyroscope panel and datasource plugins

petethepig:
Hi Grafana team!

This PR adds [pyroscope](https://github.com/pyroscope-io/pyroscope/) panel and datasource plugins.

To test the datasource plugin you can run `docker run -it -p 4040:4040 pyroscope/pyroscope:latest server` and use localhost:4040 as the address in datasource config. For app name you can use `pyroscope.server.cpu` and that be pyroscope profiling itself. We also have a publicly available server at https://demo.pyroscope.io/ .

We also have a live demo deployed at [grafana-demo.pyroscope.io](https://grafana-demo.pyroscope.io/d/9_aiZ46Gz/pyroscope-plugins?orgId=1)

Both plugins are currently not signed, I'll update this PR once we get our grafana account approved. The name of the Grafana Cloud org is `pyroscopedmitry`.

### DIFF
--- a/repo.json
+++ b/repo.json
@@ -6484,12 +6484,12 @@
       "versions": [
         {
           "version": "1.0.0",
-          "commit": "a8e6b9e577a64fb712a0cb129ac472d1c53cead9",
+          "commit": "d05ea18ed3cc5ba518052ea7258c324b85b5a9d6",
           "url": "https://github.com/pyroscope-io/grafana-panel-plugin",
           "download": {
             "any": {
               "url": "https://github.com/pyroscope-io/grafana-panel-plugin/releases/download/v1.0.0/pyroscope-panel-1.0.0.zip",
-              "md5": "27a33b0bb59dea40d20ccdf654d92148"
+              "md5": "f986fbfc5189602b844b4fffc79272b4"
             }
           }
         }
--- a/repo.json
+++ b/repo.json
@@ -3903,7 +3903,7 @@
               "md5": "e3858295a0cc0fb1be76cf0ba34474bf"
             }
           }
-        }        
+        }
       ]
     },
     {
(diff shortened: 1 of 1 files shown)


marcusolsson:
Hi, and thanks for your submission!

Unfortunately, I can't seem to access the plugin URL. Have you published the releases yet?

petethepig:
@marcusolsson Hi Marcus. My bad, the releases were in "draft" state. I fixed it now. Let us know when we can sign the plugins and I will update the PR with signed versions.

marcusolsson:
I've configured the plugins to be signed as community plugins.

Please consider providing a test configuration using the following template:

https://github.com/grafana/plugin-test-environment

petethepig:
Thank you @marcusolsson! I was able to sign the plugins so I updated this PR with signed versions.

I also set up the test environment in this repository: https://github.com/pyroscope-io/pyroscope-plugin-test-environment

Let me know if you need any help testing the plugins.

marcusolsson:
I'm getting an error sometimes when I change the Application in the query editor. I haven't found a way to reliably reproduce it, but if I keep changing the Application value, eventually it pops up.

Screenshot 2021-06-11 at 13 37 55@2x

I see that the data source is simply passing a single field-single row with a JSON body. This couples the panel and datasource so that they in practice can't be used with other plugins. I hope that you will consider finding a data frame schema that let users use the flame graph panel along with other data sources. 

Let me know if you want to look at the `s is undefined` before I publish this or if you want to fix it in the next release!

petethepig:
Hi @marcusolsson 

RE the error — we fixed the error, re-uploaded the plugin and updated this PR. I think it's good to go!

Regarding the coupling of the panel and datasource. This is a tricky problem — pyroscope returns data in a format that represents a tree. Other visualizations mostly work with arrays of some sort. So it's hard to map that tree data to arrays, while ensuring that the data visualized is useful. Having said that, we definitely see the appeal of being able to visualize pyroscope data with other visualizations. We'll think about it some more and we will probably add something in the next release.

marcusolsson:
I totally understand the challenge and it's not a blocker for publishing this. Something to keep at the back of your mind 😄 

I've published both plugins, but allow a few minutes before it's available.

PR #969, opened 2021-06-10
OPC UA Datasource Plugin v1.1.2

alydersen:
This datasource plugin allows Grafana to get data directly from one or more OPC UA Servers. Documentation in the plugin, also on connecting to a demo server


marcusolsson:
As this is has now been migrated to the Grafana organization, there's no longer a need to submit it through the community process. The Grafana team members who maintains the plugin will make sure this gets published.

PR #973, opened 2021-06-15
Updated larona-epict-panel

LucasArona:
Version 2.0.4 of the plugin, the only change is in the `plugin.json` file to allow installing it on Grafana v8.

### DIFF
--- a/repo.json
+++ b/repo.json
@@ -3853,6 +3853,16 @@
               "md5": "e22513271e3c031d44ad5870491261fd"
             }
           }
+        },
+        {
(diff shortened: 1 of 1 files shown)


marcusolsson:
I noticed that the image doesn't seem to resize to the size of the panel, which means that  it's now difficult to change the size of the panel. For smaller images, this is probably not an issue.

Screenshot 2021-07-05 at 14 10 44@2x

Since this wasn't introduced by this version, I've gone ahead and published the plugin as is.

PR #975, opened 2021-06-17
Anodot Datasource 1.0.2

gooduling:
As we don't store API token on the backend and it is stored on fronend localStorage instead, if multiple users use Grafana server, only those of them have API token stored who manually pressed buttot "Save and test" in datasource settings. This patch adds automatic requests of API token for all users of Grafana instance if right credentials were set before.


marcusolsson:
Looks great! I've published the new version 🚀

PR #990, opened 2021-07-01
Support Grafana 8 for Pyroscope Panel Plugin

Loggy:
Update pyroscope panel plugin to support Grafana 8

### DIFF
--- a/repo.json
+++ b/repo.json
@@ -2364,6 +2364,17 @@
               "md5": "bcebceb6f248194b6cc51fec118bd425"
             }
           }
+        },
+        {
+          "version": "2.3.1",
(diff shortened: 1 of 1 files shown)


marcusolsson:
The URL doesn't seem to be available. Have you published the release yet?

Also, in repo.json, could you add the version rather than updating the current 1.0.0 version? We'd like to list all the currently published versions.

Loggy:
Checked URL from different devices and browsers - looks available for me. It seems to be public and published

Also added v1.0.1 to list of versions instead of editing v1.0.0

marcusolsson:
Published! 🎉

PR #988, opened 2021-06-29
Bigquery grafana 2.0.2

ofir5300:
Already approved @daniellee 

https://github.com/grafana/grafana-plugin-repository/pull/913

PR #991, opened 2021-07-04
Bigquery grafana 2.0.2 fix

ofir5300:
A fix to #988 @daniellee 

.DS_Store file issue

### DIFF
--- a/repo.json
+++ b/repo.json
@@ -3624,7 +3624,7 @@
           "download": {
             "any": {
               "url": "https://github.com/doitintl/bigquery-grafana/releases/download/2.0.2/doitintl-bigquery-datasource-2.0.2.zip",
-              "md5": "0e13c02cea9297d8dd4c6040ffa1248d"
+              "md5": "0bb7cfe1eb1c301e52caa3416d13c8ef"
             }
           }
         }
--- a/repo.json
+++ b/repo.json
@@ -4430,7 +4430,7 @@
           "download": {
             "any": {
               "url": "https://github.com/doitintl/bigquery-grafana/releases/download/2.0.2/doitintl-bigquery-datasource-2.0.2.zip",
-              "md5": "a7c1ee373582e2264004408d0de17a4b"
+              "md5": "9f27bb6d0482cb2c72f447fd39d7fef3"
             }
           }
         }
--- a/repo.json
+++ b/repo.json
@@ -3623,8 +3623,8 @@
           "url": "https://github.com/doitintl/bigquery-grafana",
           "download": {
             "any": {
-              "url": "https://github.com/doitintl/bigquery-grafana/releases/download/2.0.2/doitintl-bigquery-datasource-2.0.2.zip",
-              "md5": "34d4aa710dd5f92da6bc155a5401a98e"
+              "url": "https://github.com/doitintl/bigquery-grafana/releases/download/2.0.2/doitintl-bigquery-datasource.zip",
+              "md5": "0e13c02cea9297d8dd4c6040ffa1248d"
             }
           }
         }
--- a/repo.json
+++ b/repo.json
@@ -4430,7 +4430,7 @@
           "download": {
             "any": {
               "url": "https://github.com/doitintl/bigquery-grafana/releases/download/2.0.2/doitintl-bigquery-datasource-2.0.2.zip",
-              "md5": "18883aceab66bc0170b87b7497aa88be"
+              "md5": "a7c1ee373582e2264004408d0de17a4b"
             }
           }
         }
--- a/repo.json
+++ b/repo.json
@@ -3624,7 +3624,7 @@
           "download": {
             "any": {
               "url": "https://github.com/doitintl/bigquery-grafana/releases/download/2.0.2/doitintl-bigquery-datasource-2.0.2.zip",
-              "md5": "ff967b1ba0bdc7a31d9153da7eddd243"
+              "md5": "18883aceab66bc0170b87b7497aa88be"
             }
           }
         }
--- a/repo.json
+++ b/repo.json
@@ -3624,7 +3624,7 @@
           "download": {
             "any": {
               "url": "https://github.com/doitintl/bigquery-grafana/releases/download/2.0.2/doitintl-bigquery-datasource-2.0.2.zip",
-              "md5": "0bb7cfe1eb1c301e52caa3416d13c8ef"
+              "md5": "ff967b1ba0bdc7a31d9153da7eddd243"
             }
           }
         }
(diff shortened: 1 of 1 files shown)


ofir5300:
@marcusolsson @daniellee 
Any news? Reminding you that a malformed v2.0.2 version is online..

daniellee:
@ofir5300 I asked Marcus to investigate the .DS_store issue - maybe we need to change something in Grafana to make this easier.

ofir5300:
Also -
When I am extracting last zip file and using the plugin folder it contains, everything run smoothly.
But when extracting the exact same zip file [downloaded from our repo](https://github.com/doitintl/bigquery-grafana/releases/download/2.0.2/doitintl-bigquery-datasource-2.0.2.zip), I am running into this error:

Screen Shot 2021-07-05 at 17 32 55
Log:

`lvl=eror msg="Failed to start plugin" logger=plugins.backend pluginId=doitintl-bigquery-datasource error="Unrecognized remote plugin message: \n\nThis usually means that the plugin is either invalid or simply\nneeds to be recompiled to support the latest protocol."`

How can it be fixed? 
Please, I don't want additional useless plugin to be approved. Users are asking about a valid update for a while now, and I started my submission 3 months ago.

Please revert [last signed version](https://grafana.com/grafana/plugins/doitintl-bigquery-datasource/) to `v2.0.1` so our plugin will be useful at the moment.

marcusolsson:
Hi @ofir5300,

I have tested the plugin with Grafana 8.0.4 running on macOS 11.2.3 (installed using brew):

- 2.0.1 from Grafana.com: `grafana-cli plugins install doitintl-bigquery-datasource 2.0.1`
- 2.0.2 from Grafana.com: `grafana-cli plugins install doitintl-bigquery-datasource 2.0.2`
- ZIP from this PR: `grafana-cli plugins --pluginUrl=https://github.com/doitintl/bigquery-grafana/releases/download/2.0.2/doitintl-bigquery-datasource-2.0.2.zip install doitintl-bigquery-datasource 2.0.2`

I've tried installing the plugin manually by downloading it and unzipping it to the plugins directory.

I've tested the same plugin versions on Grafana 8.0.4 running in Docker for Desktop on the same macOS machine.

Since the .DS_Store file is added when you open the directory in Finder, I tried doing that and got—since there's now a file in the directory that isn't included in the manifest—I got `plugin 'doitintl-bigquery-datasource' has a modified signature`, which is expected.

Unfortunately, I haven't been able to reproduce the issue you're referring to.

Would you happen to be using a Macbook with the new M1 architecture? If so, the darwin_amd64 binary won't load since it's built for the older architecture. I see that you're currently using v0.67.0 of the plugin Go SDK, and the target for `darwin_arm64` (note ARM, not AMD) was added in v.0.89.0. Would it be possible for you to update the dependency and build the binaries again? Also, if the architecture is the issue, then I don't think reverting to 2.0.1 would solve the issue.

ofir5300:
Hi @marcusolsson,
I am using Mac with intel chip, the issue I am talking about occur when extracting zip and using it on plugin directory.

I am a bit lost in your response so let's arrange it in subjects:
1. Can you be more specific about your tests, v2.0.1 worked for all envs, v2.0.2 didnt & v2.0.2 from PR raised the .DS_Store issue?
2. As far as I understand - according to you, additional binary should be added due to new arm M1 chips on macs (should compile GO files on such machine). Correct?
2. If the assumption is that architecture is NOT issue, don't you think current situation where v2.0.2 is up but cannot be used properly for all environments is bad and should be reverted? (until we figure it all out)
3. How can I fix this .DS_Store issue? I have removed it before zipping in the last zip (PR) but you are saying you ran into it again. Is it avoidable at all when extracting from zip? (I am afraid is something should be changed on your side)
4. Did you never run into this issue before? `"Unrecognized remote plugin message: \n\nThis usually means that the plugin is either invalid or simply\nneeds to be recompiled to support the latest protocol."`

My process of uploading zip is: 
Delete .DS_Store -> build & sign -> change binaries permissions (to make it executable)-> zip & upload. (anything suspicious?)

marcusolsson:
1. All my tests worked as expected. I was able to download all the plugin versions and install them on both macOS and from Docker Compose. I made an additional test where I added the .DS_Store file, which led to an error. This was expected, since it wasn't included in the manifest.
2. Correct. If you upgrade the plugin SDK to at least v0.89.0 and rebuild the backend plugin, this should create the binary for M1 macs.
3. I don't think it's necessary to revert. Since the M1 arch hasn't been supported in the past, I'd consider it a feature request for a future version.
4. Whether the .DS_Store is included or not doesn't matter to Grafana. However, since it's not really a part of the plugin, I think it should be omitted. Unfortunately, this means that if you open the plugin in Finder (which adds the .DS_Store) it has now been modified in the eyes of Grafana. In practice most users won't have to open the plugin in Finder so they won't run into this issue.
5. This was the error I was trying to reproduce, but was unable to unfortunately.

What I gather from my tests is that the currently published version is indeed working as intended. Have you received user reports to suggest otherwise, or is it mainly on your dev laptop you're seeing the issue?

Hm. I wonder whether changing the permissions after signing the plugin would affect the checksum of the files. You shouldn't need to set the permissions manually, but if you do, I think it has to be between the build and the signing steps.

If you'd like, I can republish the plugin with the version in this pull request. But since I can't reproduce the issue, I can't guarantee it fixes it.

ofir5300:
Hi @marcusolsson 
So indeed v2.0.2 works when installed properly using `grafana-cli --pluginUrl https://github.com/doitintl/bigquery-grafana/releases/download/2.0.2/doitintl-bigquery-datasource-2.0.2.zip plugins install doitintl-bigquery-datasource 2.0.2`

1. So according to you I should take both .DS_Store & unreproducible issue as edge cases which most user probably won't run into. Correct?
2. Regarding M1 chips as far as I understand it can wait to be release in future version which I think I'll submit not far from now.

Tnx much for your help & care. Of course @daniellee too!

marcusolsson:
I'm glad to hear! 💪 

1. The .DS_Store issue, yes. If you do get reports from people running into the issue I wasn't able to replicate, we'll make sure to offer expedited support and review for an update.
2. Yes! I'll expedite the review for this update as well.

Would you still like me to publish the version in this pull request, or can we close this in favor of the next version?

ofir5300:
I think we're good for now @marcusolsson 

Though I would like to try @daniellee's suggestion of using toolkit rather than webpack for my next version (https://github.com/grafana/grafana-plugin-repository/pull/913#issuecomment-865405721)

marcusolsson:
Great! Once you're able to use the grafana-toolkit, you should be able to use the recommended workflows for GitHub Actions for building, signing, and packaging your releases. I'd be happy to help you get it set up, so feel free to ping me if you run into any issues.

PR #996, opened 2021-07-08
update innius-video-panel  to v1.0.4

rvdwijngaard:
Include a few small fixes and make sure the plugin is signed.

### DIFF
--- a/repo.json
+++ b/repo.json
@@ -6130,23 +6130,13 @@
       "url": "https://github.com/innius/grafana-video-panel",
       "versions": [
         {
-          "version": "1.0.0",
-          "commit": "d4ce56ea56677143be1a6626068982766489e8d4",
-          "url": "https://github.com/innius/grafana-video-panel"
-        },
-        {
-          "version": "1.0.1",
-          "commit": "87879b0db01d13d6be6b637ad6d0e0a47f15bbb4",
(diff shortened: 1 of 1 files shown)


marcusolsson:
Looks good to me! Only one comment:

- The list of version in repo.json should list the currently published version. Unless you wish to unpublish the version, please leave them in.

In addition, I have a few suggestions for how to improve the panel but aren't necessary for publishing the plugin:

- The name of the plugin is "Video Panel". This becomes somewhat redundant when listing the visualizations (they're all panels). What do you think about change the name to just "Video"?
- In Grafana 8, the plugin description for panels is shown in the visualization list. To save space, I'd suggest changing the text from "A panel to display video from ..." to "Display video from ...".

Screenshot 2021-07-09 at 10 08 03@2x

- If you'd like the video to extend to the full width of the panel, you can remove the panel padding by setting `setNoPadding` in module.ts. Here's an [example](https://github.com/marcusolsson/grafana-calendar-panel/blob/4a2fa549207f7e7a231cb6bc4d3588fb67c07206/src/module.ts#L8).

Let me know if you'd like me to publish the plugin as is, or if you'd like to address any of the suggestions.

rvdwijngaard:
thanks for your suggestion, I made a new release which include your suggestions and support for variables.

marcusolsson:
Published! 🎉

PR #1000, opened 2021-07-12
update auxmoney-waterfall-panel to 1.0.6

cawolf:
Changelog:
* 1.0.6 (2021-07-12)
  * plugin error message and plugin data check (auxmoney/grafana-waterfall-panel#13) (bfb0555)
* 1.0.5 (2021-07-09)
  * add grafana 8 support (auxmoney/grafana-waterfall-panel#10) (a74b6d4)
* 1.0.4 (2021-04-01)
  * incorporate PR feedback for grafana/grafana-plugin-repository#863 (auxmoney/grafana-waterfall-panel#6) (2e964ea)
  * linting error (auxmoney/grafana-waterfall-panel#7) (58cb9c7)

### DIFF
--- a/repo.json
+++ b/repo.json
@@ -5907,6 +5907,16 @@
               "md5": "e90cbe4a8b4479eb21a133a9a0021566"
             }
           }
+        },
+        {
+          "version": "1.0.6",
+          "url": "https://github.com/auxmoney/grafana-waterfall-panel",
+          "download": {
+            "any": {
+              "url": "https://github.com/auxmoney/grafana-waterfall-panel/releases/download/v1.0.6/auxmoney-waterfall-panel-1.0.6.zip",
+              "md5": "d5013338b6e5755896d16c08ffc03c40"
+            }
+          }
         }
       ]
     },
@@ -6852,6 +6862,6 @@
           }
         }
       ]
-    }    
+    }
   ]
 }


marcusolsson:
Nice! I have a few suggestions, but none of them are blockers so I'd be happy to publish this as is. Let me know if you'd like to address any of them before I publish the plugin.

### Feedback

- Labels can overflow the bars if the panel is tall enough.

Screenshot 2021-07-12 at 10 02 52@2x

- The "Panel" in "Waterfall Panel" is a bit redundant when listed along all the other visualizations (they're all panels). I think just "Waterfall" would look better.

Screenshot 2021-07-12 at 10 05 32@2x

- I think "Show inline bar labels" could be shortened to "Show labels" since there's only one type of labels, right? In that case, "Show duration in bar labels" could also be shortened to "Show duration in labels".

cawolf:
Hi @marcusolsson ,

thanks for your detailed feedback. We would like to incorporate it in the next versions, so I added them as issues in our project:
* auxmoney/grafana-waterfall-panel#12
* auxmoney/grafana-waterfall-panel#14

For this version, we would like to push it out as fast as possible, as we want to enable our users to use the plugin with either Grafana 7 or 8.

marcusolsson:
Published! 🎉

PR #1009, opened 2021-07-15
Add vertica-grafana-datasource plugin

tanvipise:
Hello Grafana team,
Vertica has created a new datasource plugin with an updated framework and SDK. This plugin allows users to query and visualize data from a Vertica database.

### DIFF
--- a/repo.json
+++ b/repo.json
@@ -702,7 +702,7 @@
           "download": {
             "any": {
               "url": "https://github.com/vertica/vertica-grafana-datasource/releases/download/v2.0.0/vertica-grafana-datasource-2.0.0.zip",
-              "md5": "eff440f17920fc644b5be90e73670b49"
+              "md5": "fc59acf699889be484b56d37ca269f00"
             }
           }
         },
--- a/repo.json
+++ b/repo.json
@@ -157,8 +157,6 @@
         }
       ]
     },
-    { FOR VERTICA
-    },
     {
       "id": "aceiot-svg-panel",
       "type": "panel",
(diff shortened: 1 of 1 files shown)


briangann:
I've updated your signature type to commercial, so you can sign it during your build process.

Can you update this PR with the zip file and new md5?  Looks like you had a release at the end of June, so you might want to make a v1.0.1 just for the new signature so existing installs will see this version.

tanvipise:
Hi Brian, I commited the updated repo.json and it shows that the 'check-repo-update' has failed. Could you help with that? Thanks

briangann:
you can ignore the circleci error, it's unrelated as you mentioned. i'll fix that in another pr though.

briangann:
@tanvipise looks like the backend is using the datasource "URL" instead of the jsonData.URL parameter, and can't be provisioned from the config editor.  Using a provisioning file works, but it won't be configurable to a new install.

A provisioning file like this lets the backend connect:

```YAML
datasources:
  - name: Vertica
    version: 1
    type: vertica-grafana-datasource
    url: vertica:5433
    jsonData:
      url: vertica:5433
```

but configuring a new datasource with the config editor ends up with provisioning info like this:
```YAML
datasources:
  - name: Vertica
    version: 1
    type: vertica-grafana-datasource
    url: ""
    jsonData:
      url: vertica:5433
```

briangann (approved):
lgtm, datasource testing is working

briangann:
note: the logo.svg is not visible up on grafana.com since the fill color is set to #fff (white), and in grafana light theme it is very faint

light theme:

image

dark theme:

image

PR #1028, opened 2021-08-09
Update macropower-analytics-panel to 2.1.0

MacroPower:
v2.1.0 adds support for setting relative paths in the plugin's "endpoint" setting. Previously, only absolute paths were supported. This is an important feature for users who are running a receiver as a sidecar or in some similar setups.

To test, you can do the following:

```bash
yarn build # Or add the release to ./dist
docker-compose -f example/server/docker-compose.yaml up
```

Having a relative path isn't usable with the compose setup, but you can specify one and verify that it does send data to the proper address (and likely return a 404 error in this case).

Note that currently the docker-compose setup only works (for me) on Grafana 7. I included a bit more detail on this in #1021. Based on a few issues I believe in a Grafana 8.1.x patch this will likely be resolved. :)


marcusolsson:
Published! 🎉

